Issue #73, opened 2022-06-28
Add option to sort orchestration query results in descending order


cgillum:
There has been an ask to support querying for orchestration instances and sorting by descending order. We should investigate if this is practical at the storage provider level and, if it is, expose it all the way up to the Java client SDK.

### Suggested implementation

- [ ] Update DurableTask.Core's [OrchestrationQuery](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Query/OrchestrationQuery.cs) class to expose a new property named `SortOrder`, which is an enum with possible values of `Unspecified` (the default), `Ascending`, or `Descending`.
- [ ] Update DurableTask.AzureStorage's [OrchestrationInstanceStatusQueryCondition](https://github.com/Azure/durabletask/blob/main/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs) class to include this property and update its [ToTableQuery](https://github.com/Azure/durabletask/blob/adc96a48dbf93f3f695fcaff193180dcaa6c0386/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs#L76) method to sort the results accordingly.
- [ ] Add the sort order field to the [protobuf definition](https://github.com/microsoft/durabletask-protobuf/blob/c78f85538566fcde474ddb6d698872d1ea8d4fcf/protos/orchestrator_service.proto#L296-L305)
- [ ] Update the [Durable extension code](https://github.com/Azure/azure-functions-durable-extension/blob/7ac99a4f406ce75443d4c589f2419b08f1b58cc0/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs#L186-L198) to populate the corresponding fields when delegating query gRPC requests to the backend storage provider.
- [ ] Update the Java SDK's [OrchestrationStatusQuery](https://github.com/microsoft/durabletask-java/blob/main/client/src/main/java/com/microsoft/durabletask/OrchestrationStatusQuery.java) class with a `setSortOrder(...)` method.
- [ ] Open issues in [microsoft/durabletask-netherite](https://github.com/microsoft/durabletask-netherite) and [microsoft/durabletask-mssql](https://github.com/microsoft/durabletask-mssql) to support this functionality (I recommend adding a `help-wanted` or similar label to these issues).
- [ ] Open an issue in [microsoft/durabletask-dotnet](https://github.com/microsoft/durabletask-dotnet) to also support this functionality.

[...]

PR #74, opened 2022-06-29
add logic support reverse query results

kaibocai:
This PR add logic to support reversing query results. https://github.com/microsoft/durabletask-java/issues/73


cgillum:
After some discussion, we decided to abandon this work item. The primary issue is that Azure Storage (the most common storage backend) doesn't support ordering the results of a query, making it impossible to implement this feature efficiently.

Issue #77, opened 2022-07-28
DurableTaskGrpcClient.getInstanceMetadata does not return null when no instance found.


YusukeTobo:
Regarding the API, getInstanceMetadata should return null when there are no instances found.
https://github.com/microsoft/durabletask-java/blob/bfbd7a8a69e7467bc3d30f7831b76941530a4044/client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java#L128
```java
     * @return a metadata record that describes the orchestration instance and its execution status, or
     *         <code>null</code> if no such instance is found.
     */
    @Nullable
    public abstract OrchestrationMetadata getInstanceMetadata(String instanceId, boolean getInputsAndOutputs);
```

However, In that case, I found that DurableTaskGrpcClient.getInstanceMetadata returns DEFAULT_INSTANCE with orchestrationState_=0 (ORCHESTRATION_STATUS_RUNNING).

// Protobuf code.
```java
    /**
     * <code>.OrchestrationState orchestrationState = 2;</code>
     * @return The orchestrationState.
     */
    @java.lang.Override
    public com.microsoft.durabletask.implementation.protobuf.OrchestratorService.OrchestrationState getOrchestrationState() {
      return orchestrationState_ == null ? com.microsoft.durabletask.implementation.protobuf.OrchestratorService.OrchestrationState.getDefaultInstance() : orchestrationState_;
    }
 
 
    // @@protoc_insertion_point(class_scope:OrchestrationState)
    private static final com.microsoft.durabletask.implementation.protobuf.OrchestratorService.OrchestrationState DEFAULT_INSTANCE;
    static {
      DEFAULT_INSTANCE = new com.microsoft.durabletask.implementation.protobuf.OrchestratorService.OrchestrationState();
    }
    public static com.microsoft.durabletask.implementation.protobuf.OrchestratorService.OrchestrationState getDefaultInstance() {
      return DEFAULT_INSTANCE;
    }
```

This issue prevents us from invoking client.scheduleNewOrchestrationInstance in Singleton pattern described the official document here. That also means the library users cannot tell the accurate state of instance.

https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-singletons?tabs=java#singleton-example
```java
@FunctionName("HttpStartSingle")
public HttpResponseMessage runSingle(
        @HttpTrigger(name = "req") HttpRequestMessage<?> req,
[...]

PR #79, opened 2022-08-01
fix issue of finding instance

kaibocai:
fix https://github.com/microsoft/durabletask-java/issues/77


cgillum (approved):

cgillum:
We should also fix the JavaDoc comment since it claims we return null, which isn’t accurate.

kaibocai:
> We should also fix the JavaDoc comment since it claims we return null, which isn’t accurate.

I found update the method logic to align with the document make more sense to me. Please checkout the updated method. Thank you.

kaibocai:
> > We should also fix the JavaDoc comment since it claims we return null, which isn’t accurate.
> 
> I found update the method logic to align with the document make more sense to me. Please checkout the updated method. Thank you.

I think I am wrong, this will also required to update our sample in our official document. Will revoke the commit and update the java doc accordingly.

cgillum (approved):
LGTM!

Issue #72, opened 2022-06-24
Use middleware to replace OrchestrationRunner.loadAndRun in orchestrator functions


cgillum:
See https://github.com/Azure/azure-functions-java-worker/issues/595 for reference.

The idea is to simplify the orchestrator function programming model from this:

```java
@FunctionName("Chaining")
public String helloCitiesOrchestrator(
        @DurableOrchestrationTrigger(name = "runtimeState") String runtimeState) {
    return OrchestrationRunner.loadAndRun(runtimeState, ctx -> {
        String input = ctx.getInput(String.class);
        int x = ctx.callActivity("F1", input, int.class).await();
        int y = ctx.callActivity("F2", x, int.class).await();
        int z = ctx.callActivity("F3", y, int.class).await();
        return  ctx.callActivity("F4", z, double.class).await();
    });
}
```

...to something that looks like this...

```java
@FunctionName("Chaining")
public double helloCitiesOrchestrator(
        @DurableOrchestrationTrigger(name = "runtimeState") String input) {
    int x = ctx.callActivity("F1", input, int.class).await();
    int y = ctx.callActivity("F2", x, int.class).await();
    int z = ctx.callActivity("F3", y, int.class).await();
    return  ctx.callActivity("F4", z, double.class).await();
}
```

kaibocai:
Hi @cgillum , as far as I know, this middleware support is only available on isolated dotnet worker. I wonder for in-pro dotnent worker how durable function register the middleware? Thank you.

cgillum:
I've added this to the v1.0.0 (STABLE) milestone. The reason for prioritizing it is because it will be difficult to apply this change later.

PR #82, opened 2022-09-07
duralbe function middleware implementation

kaibocai:
This PR trying to resolve https://github.com/microsoft/durabletask-java/issues/72

This PR includes
1. a middleware that works with java worker at https://github.com/Azure/azure-functions-java-worker/tree/kaibocai/ts-middleware and java core lib at https://github.com/Azure/azure-functions-java-core-library/tree/kaibocai/middleware-support
2. a middleware class that use SPI to be loaded by java worker. 

The PR with the right worker and core lib can run the durable functions from end to end

### DIFF
--- a/azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/middleware/OrchestrationMiddleware.java
+++ b/azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/middleware/OrchestrationMiddleware.java
@@ -1,0 +1,63 @@
+package com.microsoft.durabletask.azurefunctions.middleware;
+
+import com.microsoft.azure.functions.ExecutionContext;
+import com.microsoft.azure.functions.middleware.FunctionWorkerChain;
+import com.microsoft.azure.functions.middleware.FunctionWorkerMiddleware;
+import com.microsoft.durabletask.OrchestrationRunner;
+import com.microsoft.durabletask.OrchestratorBlockedEvent;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Parameter;
+import java.util.Map;
+
+public class OrchestrationMiddleware implements FunctionWorkerMiddleware {
+    private static final String ORCHESTRATION_TRIGGER = "DurableOrchestrationTrigger";
+    private String inputName;
+    @Override
+    public void invoke(ExecutionContext context, FunctionWorkerChain next) throws Exception {
+
+        if (!isOrchestrationTrigger(context)) {
+            next.doNext(context);
+            return;
+        }
+
+        Map<String, String> argumentPayloadMap = context.getArgumentPayloadMap();
+        String base64EncodedOrchestratorRequest = argumentPayloadMap.get(this.inputName);
+
+        String orchestratorOutput  = OrchestrationRunner.loadAndRun(base64EncodedOrchestratorRequest, ctx -> {
+            try {
+                context.setInputArgument(this.inputName, ctx);
+                next.doNext(context);
+                return context.getReturnValue();
+            } catch (Exception e) {
+                if (e.getCause() instanceof OrchestratorBlockedEvent) {
+                    throw (OrchestratorBlockedEvent) e.getCause();
+                }
+                throw new RuntimeException("Unexpected failure in the task execution", e);
+            }
+        });
+
+        context.setImplicitOutput(orchestratorOutput);
+    }
+
+    private boolean isOrchestrationTrigger(ExecutionContext context) {
+        Map<String, Parameter> paramInfoMap = context.getParamInfoMap();
+        for (Map.Entry<String, Parameter> entry : paramInfoMap.entrySet()){
+            Parameter parameter = entry.getValue();
+            if (paramHasAnnotation(parameter)){
+                this.inputName = entry.getKey();
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean paramHasAnnotation(Parameter parameter) {
+        for (Annotation annotation : parameter.getAnnotations()) {
+            if (annotation.annotationType().getSimpleName().equals(ORCHESTRATION_TRIGGER)) {
+                return true;
+            }
+        }
+        return false;
+    }
+}
--- a/samples-azure-functions/src/main/java/com/functions/AzureFunctions.java
+++ b/samples-azure-functions/src/main/java/com/functions/AzureFunctions.java
@@ -4,8 +4,7 @@ import com.microsoft.azure.functions.annotation.*;
 import com.microsoft.azure.functions.*;
 import java.util.*;
 
-import com.microsoft.durabletask.DurableTaskClient;
-import com.microsoft.durabletask.OrchestrationRunner;
+import com.microsoft.durabletask.*;
 import com.microsoft.durabletask.azurefunctions.DurableActivityTrigger;
 import com.microsoft.durabletask.azurefunctions.DurableClientContext;
 import com.microsoft.durabletask.azurefunctions.DurableClientInput;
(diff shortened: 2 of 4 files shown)


azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/middleware/OrchestrationMiddleware.java
```diff
@@ -0,0 +1,63 @@
...
+
+        context.setImplicitOutput(orchestratorOutput);
+    }
+
+    private boolean isOrchestrationTrigger(ExecutionContext context) {
+        Map<String, Parameter> paramInfoMap = context.getParamInfoMap();
+        for (Map.Entry<String, Parameter> entry : paramInfoMap.entrySet()){
+            Parameter parameter = entry.getValue();
+            if (paramHasAnnotation(parameter)){
+                this.inputName = entry.getKey();
```
cgillum, line 48: Assigning an instance variable like this is dangerous from both a thread safety and a correctness/maintainability standpoint. To make the code safer, more robust, and simpler, I suggest you instead return the input name value from this method and store it as a local variable. Then, make this a `static` method.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/middleware/OrchestrationMiddleware.java
```diff
@@ -0,0 +1,63 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean paramHasAnnotation(Parameter parameter) {
```
cgillum, line 55: Can this be made static?
```suggestion
    private static boolean paramHasAnnotation(Parameter parameter) {
```

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/middleware/OrchestrationMiddleware.java
```diff
@@ -0,0 +1,63 @@
...
+import com.microsoft.azure.functions.middleware.FunctionWorkerChain;
+import com.microsoft.azure.functions.middleware.FunctionWorkerMiddleware;
+import com.microsoft.durabletask.OrchestrationRunner;
+import com.microsoft.durabletask.OrchestratorBlockedEvent;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Parameter;
+import java.util.Map;
+
+public class OrchestrationMiddleware implements FunctionWorkerMiddleware {
```
cgillum, line 13: Since this is a public class, should be add JavaDoc comments?

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/middleware/OrchestrationMiddleware.java
```diff
@@ -0,0 +1,63 @@
...
+
+        Map<String, String> argumentPayloadMap = context.getArgumentPayloadMap();
+        String base64EncodedOrchestratorRequest = argumentPayloadMap.get(this.inputName);
+
+        String orchestratorOutput  = OrchestrationRunner.loadAndRun(base64EncodedOrchestratorRequest, ctx -> {
+            try {
+                context.setInputArgument(this.inputName, ctx);
+                next.doNext(context);
+                return context.getReturnValue();
+            } catch (Exception e) {
```
cgillum, line 32: What is the purpose of this `catch` block? I wasn't expecting that we'd need anything like this. The `OrchestrationRunner.loadAndRun(...)` method should already be taking care of exception handling.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/middleware/OrchestrationMiddleware.java
```diff
@@ -0,0 +1,63 @@
+package com.microsoft.durabletask.azurefunctions.middleware;
```
cgillum, line 1: Let's add the copyright header comment to the top of this file.

samples-azure-functions/src/main/java/com/functions/AzureFunctions.java
```diff
@@ -35,17 +34,28 @@ public HttpResponseMessage startOrchestration(
...
-            @DurableOrchestrationTrigger(name = "orchestratorRequestProtoBytes") String orchestratorRequestProtoBytes) {
-        return OrchestrationRunner.loadAndRun(orchestratorRequestProtoBytes, ctx -> {
+    public String citiesOrchestratorTest(
+            @DurableOrchestrationTrigger(name = "taskOrchestrationContext") TaskOrchestrationContext ctx) throws TaskFailedException, OrchestratorBlockedEvent {
             String result = "";
             result += ctx.callActivity("Capitalize", "Tokyo", String.class).await() + ", ";
             result += ctx.callActivity("Capitalize", "London", String.class).await() + ", ";
             result += ctx.callActivity("Capitalize", "Seattle", String.class).await() + ", ";
             result += ctx.callActivity("Capitalize", "Austin", String.class).await();
             return result;
```
cgillum, line 58: It looks like these 6 lines of code need to be un-indented by 4 spaces.

samples-azure-functions/src/main/java/com/functions/AzureFunctions.java
```diff
@@ -35,17 +34,28 @@ public HttpResponseMessage startOrchestration(
...
+//        return OrchestrationRunner.loadAndRun(orchestratorRequestProtoBytes, ctx -> {
+//            String result = "";
+////            result += ctx.callActivity("Capitalize", "Tokyo", String.class).await() + ", ";
+////            result += ctx.callActivity("Capitalize", "London", String.class).await() + ", ";
+////            result += ctx.callActivity("Capitalize", "Seattle", String.class).await() + ", ";
+//            result += ctx.callActivity("Capitalize", "Austin", String.class).await();
+//            return result;
+//        });
+//    }
+
```
cgillum, line 49: It's probably best to just delete this old code rather than leaving it in commented out.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/middleware/OrchestrationMiddleware.java
```diff
@@ -0,0 +1,63 @@
...
+        Map<String, String> argumentPayloadMap = context.getArgumentPayloadMap();
+        String base64EncodedOrchestratorRequest = argumentPayloadMap.get(this.inputName);
+
+        String orchestratorOutput  = OrchestrationRunner.loadAndRun(base64EncodedOrchestratorRequest, ctx -> {
+            try {
+                context.setInputArgument(this.inputName, ctx);
+                next.doNext(context);
+                return context.getReturnValue();
+            } catch (Exception e) {
+                if (e.getCause() instanceof OrchestratorBlockedEvent) {
```
cgillum, line 33: I would expect `e instanceof OrchestratorBlockedEvent` to be true sometimes, but I'd never expect `e.getCause() instanceof OrchestratorBlockedEvent` to be true. What case are you checking for here?

kaibocai: The OrchestratorBlockedEvent throw out in cx code will be wrapped by jvm as an InvocationTargetException at https://github.com/Azure/azure-functions-java-worker/blob/ac6cb815d9493a0fcef7dd87a3006da3623b8b18/src/main/java/com/microsoft/azure/functions/worker/broker/JavaMethodInvokeInfo.java#L22, so the exception will be InvocationTargetException with cause OrchestratorBlockedEvent. I don't want to do the check inside java worker which couple the worker and sdk together. This is what I am thinking. This also apply to previous comment.

kaibocai: Yes I will update the code with comment. Besides `InvocationTargetException`, there two more exception that can be thrown out when using refection to invoke method `IllegalAccessException`, `IllegalArgumentException`, so I am using `Exception` to include all of them. 

Just wondering what's the concern here if we using `Exception`. Thanks.

cgillum: I see. In that case, `Exception` is fine. Usually, it's best practice to catch only specific exceptions, but in this case it looks like there are multiple exception types that need to be handled.

kaibocai:
new pr created https://github.com/microsoft/durabletask-java/pull/87

Issue #78, opened 2022-07-29
com.microsoft.durabletask.DataConverter$DataConverterException in case of no default constructor


YusukeTobo:
If no default constructor is generated by javac, the following deserializing warning raises in OrchestrationTrigger. The message is not clear for this library users.

```
WARNING: The orchestrator failed with an unhandled exception: com.microsoft.durabletask.DataConverter$DataConverterException: Failed to deserialize the JSON text to net.yutobo.durablefunc.Person.
```

Regarding the java debugger output, this warning is from the exception of Jackson.

```
InvalidDefinitionException@92 "com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `net.yutobo.durablefunc.Person` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: (String)"{"name":"yusuke","age":30}"; line: 1, column: 2]"
```

## Client Function
```java
        Person person = new Person("foobar", 30);
        client.scheduleNewOrchestrationInstance("orchestrateSingleton", person, instanceID);
```

## OrchestrationTrigger
```java
return OrchestrationRunner.loadAndRun(runtimeState, ctx -> {

            Person person = ctx.getInput(Person.class); // com.microsoft.durabletask.DataConverter$DataConverterException
            context.getLogger().info(person.toString());

```

kaibocai:
Hi @cgillum , I wonder what should be the fix here. Jackson cannot desterilized model that doesn't have an empty constructor and the model contains constructor with parameters which didn't annotated its parameters with @JsonProperty("field_name").

Should we update the exception message or should we write this restriction on our documents that when cx is using any class defined themselves them need to provide empty constructor?

cgillum:
@kaibocai I think the first step is to make sure this additional error information is included in the exception we raise.

> InvalidDefinitionException@92 "com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `net.yutobo.durablefunc.Person` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: (String)"{"name":"yusuke","age":30}"; line: 1, column: 2]"

Then at least developers can understand what happened without needing to use the debugger.

PR #83, opened 2022-09-13
update DataConverterException with detailed error message

kaibocai:
This PR update DataConverterException with detailed error message to provide more detailed info about the exception. 

Before update cx get warning as 
```
WARNING: The orchestrator failed with an unhandled exception: com.microsoft.durabletask.DataConverter$DataConverterException: Failed to deserialize the JSON text to net.yutobo.durablefunc.Person.
```
After update cx will get 
```
SEVERE: The orchestrator failed with an unhandled exception: com.microsoft.durabletask.DataConverter$DataConverterException: Failed to deserialize the JSON text to com.functions.model.Person. Detailed error message: Cannot construct instance of `com.functions.model.Person` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator) at [Source: (String)"{"name":"kc","age":12}"; line: 1, column: 2]
```
Resolve https://github.com/microsoft/durabletask-java/issues/78

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/JacksonDataConverter.java
+++ b/client/src/main/java/com/microsoft/durabletask/JacksonDataConverter.java
@@ -25,7 +25,7 @@ public final class JacksonDataConverter implements DataConverter {
             return jsonObjectMapper.writeValueAsString(value);
         } catch (JsonProcessingException e) {
             throw new DataConverterException(
-                    String.format("Failed to serialize argument of type '%s'.", value.getClass().getName()),
+                    String.format("Failed to serialize argument of type '%s'. Detail error message: %s", value.getClass().getName(), e.getMessage()),
                     e);
         }
     }
@@ -39,7 +39,7 @@ public final class JacksonDataConverter implements DataConverter {
         try {
             return jsonObjectMapper.readValue(jsonText, targetType);
         } catch (JsonProcessingException e) {
-            throw new DataConverterException(String.format("Failed to deserialize the JSON text to %s.", targetType.getName()), e);
+            throw new DataConverterException(String.format("Failed to deserialize the JSON text to %s. Detail error message: %s", targetType.getName(), e.getMessage()), e);
         }
     }
 }
--- a/client/src/main/java/com/microsoft/durabletask/JacksonDataConverter.java
+++ b/client/src/main/java/com/microsoft/durabletask/JacksonDataConverter.java
@@ -25,7 +25,7 @@ public final class JacksonDataConverter implements DataConverter {
             return jsonObjectMapper.writeValueAsString(value);
         } catch (JsonProcessingException e) {
             throw new DataConverterException(
-                    String.format("Failed to serialize argument of type '%s'. Detail error message: %s", value.getClass().getName(), e.getMessage()),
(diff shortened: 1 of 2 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -47,7 +47,7 @@ public TaskOrchestratorResult execute(List<HistoryEvent> pastEvents, List<Histor
         } catch (Exception e) {
             // The orchestrator threw an unhandled exception - fail it
             // TODO: What's the right way to log this?
-            logger.warning("The orchestrator failed with an unhandled exception: " + e.toString());
+            logger.severe("The orchestrator failed with an unhandled exception: " + e.toString());
```
shreyas-gopalakrishna, line 50: Double checking if it's okay to make this `severe` as if there are SLA sites, they pick these logs to report Sev 2s.

client/src/main/java/com/microsoft/durabletask/JacksonDataConverter.java
```diff
@@ -25,7 +25,7 @@ public String serialize(Object value) {
             return jsonObjectMapper.writeValueAsString(value);
         } catch (JsonProcessingException e) {
             throw new DataConverterException(
-                    String.format("Failed to serialize argument of type '%s'.", value.getClass().getName()),
+                    String.format("Failed to serialize argument of type '%s'. Detail error message: %s", value.getClass().getName(), e.getMessage()),
```
cgillum, line 28:
```suggestion
                    String.format("Failed to serialize argument of type '%s'. Detailed error message: %s", value.getClass().getName(), e.getMessage()),
```

client/src/main/java/com/microsoft/durabletask/JacksonDataConverter.java
```diff
@@ -39,7 +39,7 @@ public <T> T deserialize(String jsonText, Class<T> targetType) {
         try {
             return jsonObjectMapper.readValue(jsonText, targetType);
         } catch (JsonProcessingException e) {
-            throw new DataConverterException(String.format("Failed to deserialize the JSON text to %s.", targetType.getName()), e);
+            throw new DataConverterException(String.format("Failed to deserialize the JSON text to %s. Detail error message: %s", targetType.getName(), e.getMessage()), e);
```
cgillum, line 42:
```suggestion
            throw new DataConverterException(String.format("Failed to deserialize the JSON text to %s. Detailed error message: %s", targetType.getName(), e.getMessage()), e);
```

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
cgillum: I think of `severe` as being for unhandled exceptions in the SDK code rather than in user code. This is an exception in user code, which I think is fine for warning. Remember too that the Functions host has its own logging for functions that fail with an unhandled exception.

cgillum (approved):
I added a few minor comments, but otherwise I'm good with this PR.

PR #84, opened 2022-09-13
waitForOrchestrationStart/Complete methods throw TimeoutException when grpc deadline exceeded

### LINKED ISSUE #62: The waitForOrchestrationStart/Complete methods should throw TimeoutException
Currently these methods will raise a `io.grpc.StatusRuntimeException` if the timeout deadline is exceeded.

> io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 0.998280100s. [closed=[], open=[[remote_addr=/127.0.0.1:4001]]]

This is problematic because we don't want the details of gRPC to bleed out into the programming model, particularly for things that are "expected".

Instead, we should catch the gRPC exception internally, validate that it's a `DEADLINE_EXCEEDED` exception, and then throw a `TimeoutException`. This is a checked exception, which means that callers will be required to catch it.

Note that this is a breaking change and should be done before the `1.0.0` release.

### LINKED ISSUE #64: Add waitForCompletionOrCreateCheckStatusResponse client API
Other languages support a client API that allows you to wait for an orchestration to complete and either a) return the output if it completes within the specified timeout or b) return an HTTP 202 response if the timeout expires. The first release of Java doesn't support this, although it's possible to implement manually.

More information on this feature can be found [here](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management?tabs=csharp#wait-for-orchestration-completion).

This is an Azure Functions-specific API that will need to be made to the `durabletask-azure-functions` package.

Once the API exists, the documentation linked above will need to have its sample updated to reference it.

shreyas-gopalakrishna:
- waitForOrchestrationStart/Complete methods throw TimeoutException when grpc deadline exceeded
- added unit tests
- updated docs 

resolves #62

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
@@ -162,7 +162,7 @@ final class DurableTaskGrpcClient extends DurableTaskClient {
             response = grpcClient.waitForInstanceStart(request);
         } catch (StatusRuntimeException e) {
             if (e.getStatus().getCode() == Status.Code.DEADLINE_EXCEEDED) {
-                throw new TimeoutException("Start orchestration timeout reached");
+                throw new TimeoutException("Start orchestration timeout reached.");
             }
             throw e;
         }
@@ -189,7 +189,7 @@ final class DurableTaskGrpcClient extends DurableTaskClient {
             response = grpcClient.waitForInstanceCompletion(request);
         } catch (StatusRuntimeException e) {
             if (e.getStatus().getCode() == Status.Code.DEADLINE_EXCEEDED) {
-                throw new TimeoutException("Orchestration instance completion timeout reached");
+                throw new TimeoutException("Orchestration instance completion timeout reached.");
             }
             throw e;
         }
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
@@ -15,6 +15,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Logger;
 
 /**
@@ -142,7 +143,7 @@ final class DurableTaskGrpcClient extends DurableTaskClient {
     }
 
     @Override
-    public OrchestrationMetadata waitForInstanceStart(String instanceId, Duration timeout, boolean getInputsAndOutputs) {
+    public OrchestrationMetadata waitForInstanceStart(String instanceId, Duration timeout, boolean getInputsAndOutputs) throws TimeoutException {
         GetInstanceRequest request = GetInstanceRequest.newBuilder()
                 .setInstanceId(instanceId)
                 .setGetInputsAndOutputs(getInputsAndOutputs)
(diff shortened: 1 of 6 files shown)


kaibocai:
This PR also resolves https://github.com/microsoft/durabletask-java/issues/62 right?

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
```diff
@@ -778,4 +788,54 @@ void purgeInstanceFilter() {
...
     }
+
+    @Test()
+    void waitForInstanceStartThrowsException() {
+        final String orchestratorName = "orchestratorName";
+
+        DurableTaskGrpcWorker worker = this.createWorkerBuilder()
+                .addOrchestrator(orchestratorName, ctx -> {
+                    try {
+                        TimeUnit.SECONDS.sleep(5);
```
cgillum, line 799: It might be worth adding a comment here explaining why this test is correct.
```suggestion
                        // The orchestration remains in the "Pending" state until the first await statement
                        TimeUnit.SECONDS.sleep(5);
```

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
```diff
@@ -155,12 +156,21 @@ public OrchestrationMetadata waitForInstanceStart(String instanceId, Duration ti
...
                 timeout.toMillis(),
                 TimeUnit.MILLISECONDS);
-        GetInstanceResponse response = grpcClient.waitForInstanceStart(request);
+
+        GetInstanceResponse response;
+        try {
+            response = grpcClient.waitForInstanceStart(request);
+        } catch (StatusRuntimeException e) {
+            if (e.getStatus().getCode() == Status.Code.DEADLINE_EXCEEDED) {
+                throw new TimeoutException("Start orchestration timeout reached");
```
cgillum, line 165: I'm not sure if there's a Java convention for this, but I usually try to end simple exception messages like this with a period.
```suggestion
                throw new TimeoutException("Start orchestration timeout reached.");
```

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
```diff
@@ -173,7 +183,16 @@ public OrchestrationMetadata waitForInstanceCompletion(String instanceId, Durati
...
                 timeout.toMillis(),
                 TimeUnit.MILLISECONDS);
-        GetInstanceResponse response = grpcClient.waitForInstanceCompletion(request);
+
+        GetInstanceResponse response;
+        try {
+            response = grpcClient.waitForInstanceCompletion(request);
+        } catch (StatusRuntimeException e) {
+            if (e.getStatus().getCode() == Status.Code.DEADLINE_EXCEEDED) {
+                throw new TimeoutException("Orchestration instance completion timeout reached");
```
cgillum, line 192: Same comment as before:
```suggestion
                throw new TimeoutException("Orchestration instance completion timeout reached.");
```

cgillum (approved):
LGTM! No blockers from my POV. Just a few very minor suggestions.

cgillum:
> This PR also resolves #62 right?

Yes, I think the wrong issue was linked in the description. @shreyas-gopalakrishna can you update the PR description to ensure that we're linking to the correct issue?

kaibocai (approved):

kaibocai:
let's merge this if everything is done.

PR #86, opened 2022-09-16
Throw CompositeTaskFailedException for context.allOf

### LINKED ISSUE #54: TaskOrchestrationContext.allOf should throw TaskFailureException instead of RuntimeException
The implementation of this method in TaskOrchestrationExecutor.java currently throws a `RuntimeException` if any of the inner tasks fails. This seems wrong because tasks are only supposed to fail with `TaskFailedException`.

One challenge is that `TaskFailedException` is designed to represent a single named task. We may need to create a new exception type for composite exceptions. Alternatively, we could surface the details of the first exception. It may be useful to use the current .NET implementation design as a reference for this design.

As part of this work item, we should also do the following:
- [ ] Update the API documentation for `TaskOrchestrationContext.allOf` to accurately reflect the expected exception behavior
- [ ] Write a unit or integration test that ensures the correct behavior

shreyas-gopalakrishna:
TaskOrchestrationContext.allOf should throw TaskFailureException instead of RuntimeException
- code changes 
- unit tests 
- documentation changes

resolves #54 

### Pull request checklist

    * [x] Otherwise: Documentation issue linked to PR

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
+++ b/client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
@@ -49,7 +49,7 @@ public class CompositeTaskFailedException extends RuntimeException {
      *
      * @return a list of exceptions
      */
-    List<Exception> getExceptions() {
+    public List<Exception> getExceptions() {
         return new ArrayList<>(this.exceptions);
     }
 
--- a/client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
+++ b/client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
@@ -1,40 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
 package com.microsoft.durabletask;
 
 import java.util.ArrayList;
 import java.util.List;
 
-public class CompositeTaskFailedException extends Exception{
+/**
+ * Exception that gets thrown when multiple {@link Task}s for an activity or sub-orchestration fails with an
+ * unhandled exception.
+ * <p>
+ * Detailed information associated with each task failure can be retrieved using the {@link #getExceptions()}
+ * method.
+ */
+public class CompositeTaskFailedException extends RuntimeException {
     private final List<Exception> exceptions;
 
-    public CompositeTaskFailedException() {
+    CompositeTaskFailedException() {
         this.exceptions = new ArrayList<>();
     }
 
-    public CompositeTaskFailedException(List<Exception> exceptions) {
+    CompositeTaskFailedException(List<Exception> exceptions) {
         this.exceptions = exceptions;
     }
 
-    public CompositeTaskFailedException(String message, List<Exception> exceptions) {
+    CompositeTaskFailedException(String message, List<Exception> exceptions) {
         super(message);
         this.exceptions = exceptions;
     }
 
-    public CompositeTaskFailedException(String message, Throwable cause, List<Exception> exceptions) {
+    CompositeTaskFailedException(String message, Throwable cause, List<Exception> exceptions) {
         super(message, cause);
         this.exceptions = exceptions;
     }
 
-    public CompositeTaskFailedException(Throwable cause, List<Exception> exceptions) {
+    CompositeTaskFailedException(Throwable cause, List<Exception> exceptions) {
         super(cause);
         this.exceptions = exceptions;
     }
 
-    public CompositeTaskFailedException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace, List<Exception> exceptions) {
+    CompositeTaskFailedException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace, List<Exception> exceptions) {
         super(message, cause, enableSuppression, writableStackTrace);
         this.exceptions = exceptions;
     }
 
-    public List<Exception> getExceptions() {
+    /**
+     * Gets a list of exceptions that occurred during execution of a group of {@link Task}
+     * These exceptions include details of the task failure and exception information
+     *
+     * @return a list of exceptions
+     */
+    List<Exception> getExceptions() {
         return new ArrayList<>(this.exceptions);
     }
 
--- a/client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
+++ b/client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
@@ -1,0 +1,41 @@
+package com.microsoft.durabletask;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CompositeTaskFailedException extends Exception{
+    private final List<Exception> exceptions;
+
+    public CompositeTaskFailedException() {
+        this.exceptions = new ArrayList<>();
+    }
+
+    public CompositeTaskFailedException(List<Exception> exceptions) {
+        this.exceptions = exceptions;
+    }
+
+    public CompositeTaskFailedException(String message, List<Exception> exceptions) {
+        super(message);
+        this.exceptions = exceptions;
+    }
+
+    public CompositeTaskFailedException(String message, Throwable cause, List<Exception> exceptions) {
+        super(message, cause);
+        this.exceptions = exceptions;
+    }
+
+    public CompositeTaskFailedException(Throwable cause, List<Exception> exceptions) {
+        super(cause);
+        this.exceptions = exceptions;
+    }
+
+    public CompositeTaskFailedException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace, List<Exception> exceptions) {
+        super(message, cause, enableSuppression, writableStackTrace);
+        this.exceptions = exceptions;
+    }
+
+    public List<Exception> getExceptions() {
+        return new ArrayList<>(this.exceptions);
+    }
+
+}
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
@@ -81,8 +81,8 @@ public interface TaskOrchestrationContext {
      * <pre>{@code
      * try {
      *     List<String> orderedResults = ctx.allOf(List.of(t1, t2, t3)).await();
-     * } catch{
-     *     e.getExceptions()
+     * } catch (CompositeTaskFailedException e) {
+     *     List<Exception> exceptions = e.getExceptions()
      * }
      * }</pre>
      *
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
@@ -80,7 +80,7 @@ public interface TaskOrchestrationContext {
      * @param <V> the return type of the {@code Task} objects
      * @return the values of the completed {@code Task} objects in the same order as the source list
      */
-    <V> Task<List<V>> allOf(List<Task<V>> tasks) throws CompositeTaskFailedException;
+    <V> Task<List<V>> allOf(List<Task<V>> tasks);
 
     /**
      * Returns a new {@code Task} that is completed when any of the tasks in {@code tasks} completes.
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
@@ -61,7 +61,7 @@ public interface TaskOrchestrationContext {
     //       https://github.com/microsoft/durabletask-java/issues/54
     /**
      * Returns a new {@code Task} that is completed when all the given {@code Task}s complete. If any of the given
-     * {@code Task}s complete with an exception, the returned {@code Task} will also complete with an exception
+     * {@code Task}s complete with an exception, the returned {@code Task} will also complete with an {@link CompositeTaskFailedException}
      * containing details of the first encountered failure. The value of the returned {@code Task} is an ordered list of
      * the return values of the given tasks. If no tasks are provided, returns a {@code Task} completed with value
      * {@code null}.
@@ -74,6 +74,16 @@ public interface TaskOrchestrationContext {
      * Task<String> t3 = ctx.callActivity("MyActivity", String.class);
      *
      * List<String> orderedResults = ctx.allOf(List.of(t1, t2, t3)).await();
+     * }</pre>
+     *
+     * Exceptions in any of the given tasks results in an unchecked {@link CompositeTaskFailedException}.
+     * This exception can be inspected to obtain failure details of individual {@link Task}s.
+     * <pre>{@code
+     * try {
+     *     List<String> orderedResults = ctx.allOf(List.of(t1, t2, t3)).await();
+     * } catch{
+     *     e.getExceptions()
+     * }
      * }</pre>
      *
      * @param tasks the list of {@code Task} objects
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
@@ -80,7 +80,7 @@ public interface TaskOrchestrationContext {
      * @param <V> the return type of the {@code Task} objects
      * @return the values of the completed {@code Task} objects in the same order as the source list
      */
-    <V> Task<List<V>> allOf(List<Task<V>> tasks);
+    <V> Task<List<V>> allOf(List<Task<V>> tasks) throws CompositeTaskFailedException;
 
     /**
      * Returns a new {@code Task} that is completed when any of the tasks in {@code tasks} completes.
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
@@ -180,7 +180,7 @@ final class TaskOrchestrationExecutor {
                             try {
                                 results.add(cf.get());
                             } catch (Exception ex) {
-                                throw new RuntimeException("One or more tasks failed.", ex);
+                                results.add(null);
                             }
                         }
                         return results;
@@ -194,7 +194,7 @@ final class TaskOrchestrationExecutor {
                                 exceptions.add(ex);
                             }
                         }
-                        throw new CompositeTaskFailedException("One or more tasks failed.", exceptions);
+                        throw new CompositeTaskFailedException(exceptions.get(0).getCause().getMessage(), exceptions);
                     })
             );
         }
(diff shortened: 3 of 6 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -166,30 +164,39 @@ public <V> Task<V> completedTask(V value) {
...
                     } catch (Exception ex) {
-                        // TODO: Better exception message than this
-                        // TODO: This needs to be a TaskFailedException or some other documented exception type.
-                        //       https://github.com/microsoft/durabletask-java/issues/54
-                        throw new RuntimeException("One or more tasks failed.", ex);
+                        exceptions.add(ex);
                     }
                 }
 
+                if(exceptions.size() > 0){
```
shreyas-gopalakrishna, line 189: @cgillum @kaibocai the thought was to catch exceptions into a list and use that to throw a CompositeTaskFailedException but we cannot throw it from here. Any ideas how we could handle this?

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -166,30 +164,39 @@ public <V> Task<V> completedTask(V value) {
...
-        public <V> Task<List<V>> allOf(List<Task<V>> tasks) {
+        public <V> Task<List<V>> allOf(List<Task<V>> tasks) throws CompositeTaskFailedException{
             Helpers.throwIfArgumentNull(tasks, "tasks");
 
             CompletableFuture<V>[] futures = tasks.stream()
                     .map(t -> t.future)
                     .toArray((IntFunction<CompletableFuture<V>[]>) CompletableFuture[]::new);
 
-            return new CompletableTask<>(CompletableFuture.allOf(futures).thenApply(x -> {
+//            List<Exception> exceptions = new ArrayList<>(futures.length);
```
shreyas-gopalakrishna, line 174: exceptions cannot be collected here as I think all futures wouldn't be completed here yet.

kaibocai: Sorry I don't get it, why can't we move the `exceptions` list out at this place? It doesn't matter if the futures completed or not when we initialize the exceptions list right.

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
```diff
@@ -778,4 +783,41 @@ void purgeInstanceFilter() {
...
+                            .mapToObj(i -> ctx.callActivity(activityName, i, Integer.class))
+                            .collect(Collectors.toList());
+
+                    // Wait for all tasks to complete
+                    List<Integer> results;
+                    try {
+                        results = ctx.allOf(parallelTasks).await();
+                        ctx.complete(results);
+                    } catch (CompositeTaskFailedException e) {
+                        e.printStackTrace();
```
cgillum, line 807: This test verifies that the correct exception type was thrown but doesn't verify that it's actually a composite containing details about each of the individual failures. We should make this test more comprehensive and actually inspect the contents of the exception.

shreyas-gopalakrishna: Yes, I am yet to add that since CompositeTaskFailedException is current not being raised here and second catch block catches the Exception. 

image

client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.List;
+
+public class CompositeTaskFailedException extends Exception{
+    private final List<Exception> exceptions;
+
+    public CompositeTaskFailedException() {
+        this.exceptions = new ArrayList<>();
+    }
+
+    public CompositeTaskFailedException(List<Exception> exceptions) {
```
cgillum, line 13: Let's make all these constructors package-internal (remove `public`). We don't want to add more to the public API surface area than absolutely necessary.

client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
```diff
@@ -0,0 +1,41 @@
+package com.microsoft.durabletask;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CompositeTaskFailedException extends Exception{
```
cgillum, line 6: Let's add JavaDoc comments to all the public APIs. Please refer to the existing JavaDocs in this code as examples for style guidelines.

client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
```diff
@@ -0,0 +1,41 @@
+package com.microsoft.durabletask;
```
cgillum, line 1: Please add the license header for new files.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
cgillum: Correct. You can't throw it from here. It has to be thrown from the `.await()` call of the `Task<List<V>>` object that we return from this method - i.e. `completableTask`.  You may need to set some state in this object somehow so that we know what to throw.

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
```diff
@@ -317,7 +317,7 @@ void termination() {
     }
 
     @Test
-    void activityFanOut() throws IOException {
+    void activityFanOut() throws TaskFailedException {
```
cgillum, line 320: Why do we need this change?

samples/src/main/java/io/durabletask/samples/FanOutFanInPattern.java
```diff
@@ -70,7 +70,12 @@ public TaskOrchestration create() {
                             .collect(Collectors.toList());
 
                     // Fan-in to get the total word count from all the individual activity results.
-                    List<Integer> allWordCountResults = ctx.allOf(tasks).await();
+                    List<Integer> allWordCountResults;
+                    try {
+                        allWordCountResults = ctx.allOf(tasks).await();
+                    } catch (CompositeTaskFailedException e) {
```
cgillum, line 76: Since you had to change this sample, it means we're making a breaking change. I wonder if we should instead make `CompositeTaskFailedException` a runtime exception to avoid the breaking change. Alternatively, we should consider updating `TaskOrchestration.run` to catch this exception so that developers don't have to.

cgillum (commented):
Some thoughts

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
cgillum: Something to consider: when designing and implementing new APIs, it’s a best practice to write the test first before completing the implementation. In this way, you’re designing for the customer ideal experience since the test represents the customer experience. This is the spirit behind the TDD movement. I recommend looking into this practice moving forward.

client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+/**
+ * Exception that gets thrown when multiple {@link Task}s for an activity or sub-orchestration fails with an
+ * unhandled exception.
+ * <p>
+ * Detailed information associated with each task failure can be retrieved using the {@link #getExceptions()}
+ * method.
+ */
+public class CompositeTaskFailedException extends RuntimeException {
```
kaibocai, line 15: Just to check if here is uncheck exception then when cx using `allOf` method they won't know to catch the specific CompositeTaskFailedException right. I believe we will document this for `allOf` method somewhere. 
Also should we update the API javadoc for `allOf` mehtod.

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
```diff
@@ -778,4 +779,54 @@ void purgeInstanceFilter() {
...
+            assertEquals(OrchestrationRuntimeStatus.FAILED, instance.getRuntimeStatus());
+
+            List<?> output = instance.readOutputAs(List.class);
+            assertNull(output);
+
+            FailureDetails details = instance.getFailureDetails();
+            assertNotNull(details);
+            assertEquals("One or more tasks failed.", details.getErrorMessage());
+            assertEquals("com.microsoft.durabletask.CompositeTaskFailedException", details.getErrorType());
+            assertNotNull(details.getStackTrace());
```
kaibocai, line 829: We can also add a exception size assert to double check here. Just to double check :)

shreyas-gopalakrishna: Yes, this is done in line 802

client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
shreyas-gopalakrishna: I can document it as part of the Javadoc.

kaibocai (approved):

client/src/main/java/com/microsoft/durabletask/CompositeTaskFailedException.java
```diff
@@ -0,0 +1,56 @@
...
+        this.exceptions = exceptions;
+    }
+
+    /**
+     * Gets a list of exceptions that occurred during execution of a group of {@link Task}
+     * These exceptions include details of the task failure and exception information
+     *
+     * @return a list of exceptions
+     */
+    List<Exception> getExceptions() {
```
cgillum, line 52: This needs to be public.

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
```diff
@@ -778,4 +779,54 @@ void purgeInstanceFilter() {
...
+                        ctx.complete(results);
+                    }catch (CompositeTaskFailedException e){
+                        assertNotNull(e);
+                        assertEquals(2, e.getExceptions().size());
+                        assertEquals(ExecutionException.class, e.getExceptions().get(0).getClass());
+                        assertEquals(ExecutionException.class, e.getExceptions().get(1).getClass());
+                        assertEquals(TaskFailedException.class, e.getExceptions().get(0).getCause().getClass());
+                        assertEquals(TaskFailedException.class, e.getExceptions().get(1).getCause().getClass());
+                        assertTrue(e.getExceptions().get(0).getCause().getMessage().contains("unhandled exception: / by zero"));
+                        assertTrue(e.getExceptions().get(1).getCause().getMessage().contains("unhandled exception: / by zero"));
```
cgillum, line 808: I have concerns about what the full error message will look like. Can we update the test to do an `.equals` check instead of `.contains`?

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
```diff
@@ -778,4 +779,54 @@ void purgeInstanceFilter() {
...
+            OrchestrationMetadata instance = client.waitForInstanceCompletion(instanceId, defaultTimeout, true);
+            assertNotNull(instance);
+            assertEquals(OrchestrationRuntimeStatus.FAILED, instance.getRuntimeStatus());
+
+            List<?> output = instance.readOutputAs(List.class);
+            assertNull(output);
+
+            FailureDetails details = instance.getFailureDetails();
+            assertNotNull(details);
+            assertEquals("One or more tasks failed.", details.getErrorMessage());
```
cgillum, line 827: This doesn't seem very helpful. Can we include more information, like which task failed (using the task name and/or ID if available)?

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -173,23 +171,32 @@ public <V> Task<List<V>> allOf(List<Task<V>> tasks) {
...
+            return new CompletableTask<>(CompletableFuture.allOf(futures)
+                    .thenApply(x -> {
+                        List<V> results = new ArrayList<>(futures.length);
+
+                        // All futures are expected to be completed at this point
+                        for (CompletableFuture<V> cf : futures) {
+                            try {
+                                results.add(cf.get());
+                            } catch (Exception ex) {
+                                throw new RuntimeException("One or more tasks failed.", ex);
```
cgillum, line 183: Do we still need this catch/rethrow here if we have the `exceptionally` handler? I don't understand why we would throw this exception here with the exact same text as the `CompositeTaskFailedException` on line 197.

shreyas-gopalakrishna: `cf.get()` throws checked exceptions, we will have to catch that. We do not need the rethrow as exceptions will be handled by `exceptionally`. I'm thinking we just log exception details in the catch block, is that good?

cgillum: No, I don't think we should log individual task failures inside this method. Those exceptions should be getting logged elsewhere already. Instead, I think we need to call `results.add(...)` with a default/null value.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
```diff
@@ -76,6 +76,16 @@ public interface TaskOrchestrationContext {
      * List<String> orderedResults = ctx.allOf(List.of(t1, t2, t3)).await();
      * }</pre>
      *
+     * Exceptions in any of the given tasks results in an unchecked {@link CompositeTaskFailedException}.
+     * This exception can be inspected to obtain failure details of individual {@link Task}s.
+     * <pre>{@code
+     * try {
+     *     List<String> orderedResults = ctx.allOf(List.of(t1, t2, t3)).await();
+     * } catch{
+     *     e.getExceptions()
```
cgillum, line 85:
```suggestion
     *     List<Exception> exceptions = e.getExceptions()
```

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
```diff
@@ -76,6 +76,16 @@ public interface TaskOrchestrationContext {
      * List<String> orderedResults = ctx.allOf(List.of(t1, t2, t3)).await();
      * }</pre>
      *
+     * Exceptions in any of the given tasks results in an unchecked {@link CompositeTaskFailedException}.
+     * This exception can be inspected to obtain failure details of individual {@link Task}s.
+     * <pre>{@code
+     * try {
+     *     List<String> orderedResults = ctx.allOf(List.of(t1, t2, t3)).await();
+     * } catch{
```
cgillum, line 84:
```suggestion
     * } catch (CompositeTaskFailedException e) {
```

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
```diff
@@ -778,4 +779,62 @@ void purgeInstanceFilter() {
...
+
+                    // Wait for all tasks to complete
+                    try {
+                        List<Integer> results = ctx.allOf(parallelTasks).await();
+                        ctx.complete(results);
+                    }catch (CompositeTaskFailedException e){
+                        assertNotNull(e);
+                        assertEquals(2, e.getExceptions().size());
+                        assertEquals(ExecutionException.class, e.getExceptions().get(0).getClass());
+                        assertEquals(ExecutionException.class, e.getExceptions().get(1).getClass());
```
cgillum, line 804: What is an `ExecutionException`?  I would have expected the type of `e.getException()` to be `TaskFailedException`. Is it possible to do this? I think that would make for a better developer experience because the `TaskFailedException` is the most interesting thing for code to inspect.

cgillum (commented):
I think we're very close on this one! There are a couple changes we should make to the JavaDoc comments and one suggested change about the organization of the inner exceptions.

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
shreyas-gopalakrishna: CompletableFuture.get() here is throwing the ExecutionException. We will have to create a new TaskFailedException but do not have all failure details like below.
```java
public <V> Task<List<V>> allOf(List<Task<V>> tasks) {
      Helpers.throwIfArgumentNull(tasks, "tasks");

      CompletableFuture<V>[] futures = tasks.stream()
              .map(t -> t.future)
              .toArray((IntFunction<CompletableFuture<V>[]>) CompletableFuture[]::new);

      return new CompletableTask<>(CompletableFuture.allOf(futures)
              .thenApply(x -> {
                  List<V> results = new ArrayList<>(futures.length);

                  // All futures are expected to be completed at this point
                  for (CompletableFuture<V> cf : futures) {
                      try {
                          results.add(cf.get());
                      } catch (Exception ex) {
                          results.add(null);
                      }
                  }
                  return results;
              })
              .exceptionally(throwable -> {
                  ArrayList<Exception> exceptions = new ArrayList<>(futures.length);
                  for (CompletableFuture<V> cf : futures) {
                      try {
                          cf.get(); // This throws ExecutionException
                      } catch (Exception ex) {
                          // We can create a TaskFailedException here but we will have to parse the exception for taskId, name and other details                           
                         exceptions.add(ex);
                      }
                  }
                  throw new CompositeTaskFailedException(exceptions.get(0).getCause().getMessage(), exceptions);
              })
      );
  }
```



image

cgillum: I see, so `CompletableFuture.get()` always throws this. Then instead of adding `ExecutionException` to the list of exceptions, can we instead just add the inner `TaskFailedException` directly? Something like this:

```java
try {
    cf.get();
} catch (ExecutionException ex) {
    // The cause is expected to be TaskFailedException
    exceptions.add(ex.getCause());
} catch (Exception ex) {
    // Not expected, but just in case we see an exception we don't know about
    exceptions.add(ex);
}
```

That way we don't need to expose `ExecutionException` in user code.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -173,23 +171,34 @@ public <V> Task<List<V>> allOf(List<Task<V>> tasks) {
...
+                        for (CompletableFuture<V> cf : futures) {
+                            try {
+                                cf.get();
+                            } catch (ExecutionException ex) {
+                                exceptions.add((Exception) ex.getCause());
+                            } catch (Exception ex){
+                                exceptions.add(ex);
+                            }
+                        }
+                        throw new CompositeTaskFailedException(exceptions.get(0).getMessage(), exceptions);
```
cgillum, line 199: I wonder if it would be better to have a more generic message here. Otherwise, developers may not realize that there could be more than one failure. How about something like this?
```suggestion
                        throw new CompositeTaskFailedException(
                            String.format(
                                "%d out of %d tasks failed with an exception. See the exceptions list for details.", 
                                exceptions.size(),
                                futures.length),
                            exceptions);
```

cgillum (commented):
Sorry, one more thing I noticed in my final pass through the code.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
shreyas-gopalakrishna: My thought was to inform the developers with the first exception message. But this message is more informative.

cgillum (approved):
This looks great! Thanks for doing this work!

PR #90, opened 2022-10-12
Updated PurgeInstances to throw TimeoutException

### LINKED ISSUE #37: The purgeInstances API should take a timeout parameter
Purging lots of orchestration data can take a very long time. We should add a timeout parameter that allows the caller to limit how much time is spent purging instance data. If the timeout expires, ideally, we should still return the number of instances purged, but possibly return an addition value in the `PurgeResult` class that indicates whether the operation was cancelled.

shreyas-gopalakrishna:
Updated PurgeInstances to take a timeout parameter and throw TimeoutException
resolves #37 

### Pull request checklist

    * [x] Otherwise: Documentation issue linked to PR

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
@@ -252,13 +252,31 @@ final class DurableTaskGrpcClient extends DurableTaskClient {
     }
 
     @Override
-    public PurgeResult purgeInstances(PurgeInstanceCriteria purgeInstanceCriteria) {
+    public PurgeResult purgeInstances(PurgeInstanceCriteria purgeInstanceCriteria) throws TimeoutException {
         PurgeInstanceFilter.Builder builder = PurgeInstanceFilter.newBuilder();
         builder.setCreatedTimeFrom(DataConverter.getTimestampFromInstant(purgeInstanceCriteria.getCreatedTimeFrom()));
         Optional.ofNullable(purgeInstanceCriteria.getCreatedTimeTo()).ifPresent(createdTimeTo -> builder.setCreatedTimeTo(DataConverter.getTimestampFromInstant(createdTimeTo)));
         purgeInstanceCriteria.getRuntimeStatusList().forEach(runtimeStatus -> Optional.ofNullable(runtimeStatus).ifPresent(status -> builder.addRuntimeStatus(OrchestrationRuntimeStatus.toProtobuf(status))));
-        PurgeInstancesResponse response = this.sidecarClient.purgeInstances(PurgeInstancesRequest.newBuilder().setPurgeInstanceFilter(builder).build());
-        return toPurgeResult(response);
+
+        Duration timeout = purgeInstanceCriteria.getTimeout();
+        if (timeout == null || timeout.isNegative() || timeout.isZero()) {
+            timeout = Duration.ofMinutes(10);
+        }
+
+        TaskHubSidecarServiceBlockingStub grpcClient = this.sidecarClient.withDeadlineAfter(
+                timeout.toMillis(),
+                TimeUnit.MILLISECONDS);
+
+        PurgeInstancesResponse response;
+        try {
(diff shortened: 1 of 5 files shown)


client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
```diff
@@ -252,13 +252,31 @@ public PurgeResult purgeInstance(String instanceId) {
...
         PurgeInstanceFilter.Builder builder = PurgeInstanceFilter.newBuilder();
         builder.setCreatedTimeFrom(DataConverter.getTimestampFromInstant(purgeInstanceCriteria.getCreatedTimeFrom()));
         Optional.ofNullable(purgeInstanceCriteria.getCreatedTimeTo()).ifPresent(createdTimeTo -> builder.setCreatedTimeTo(DataConverter.getTimestampFromInstant(createdTimeTo)));
         purgeInstanceCriteria.getRuntimeStatusList().forEach(runtimeStatus -> Optional.ofNullable(runtimeStatus).ifPresent(status -> builder.addRuntimeStatus(OrchestrationRuntimeStatus.toProtobuf(status))));
-        PurgeInstancesResponse response = this.sidecarClient.purgeInstances(PurgeInstancesRequest.newBuilder().setPurgeInstanceFilter(builder).build());
-        return toPurgeResult(response);
+
+        Duration timeout = purgeInstanceCriteria.getTimeout();
+        if (timeout == null || timeout.isNegative() || timeout.isZero()) {
+            timeout = Duration.ofMinutes(10);
```
shreyas-gopalakrishna, line 263: @cgillum checking if 10 minutes default timeout is good here

cgillum: The default timeout for Azure Functions execution is 5-minutes, so I think it would be best if we set our default to something less than this. How about 4 minutes?

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
```diff
@@ -252,13 +252,31 @@ public PurgeResult purgeInstance(String instanceId) {
...
+        TaskHubSidecarServiceBlockingStub grpcClient = this.sidecarClient.withDeadlineAfter(
+                timeout.toMillis(),
+                TimeUnit.MILLISECONDS);
+
+        PurgeInstancesResponse response;
+        try {
+            response = grpcClient.purgeInstances(PurgeInstancesRequest.newBuilder().setPurgeInstanceFilter(builder).build());
+            return toPurgeResult(response);
+        } catch (StatusRuntimeException e) {
+            if(e.getStatus().getCode() == Status.Code.DEADLINE_EXCEEDED){
```
cgillum, line 275: nit: whitespace
```suggestion
            if (e.getStatus().getCode() == Status.Code.DEADLINE_EXCEEDED) {
```

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
```diff
@@ -252,13 +252,31 @@ public PurgeResult purgeInstance(String instanceId) {
...
+                throw new TimeoutException("Purge instances completion timeout reached.");
```
cgillum, line 276: Let's include the configured timeout duration in the exception message. This will help customers understand what the timeout value is, particularly when they are using default values.

client/src/main/java/com/microsoft/durabletask/PurgeInstanceCriteria.java
```diff
@@ -58,6 +60,17 @@ public PurgeInstanceCriteria setRuntimeStatusList(List<OrchestrationRuntimeStatu
         return this;
     }
 
+    /**
+     * Purge orchestration instances within the specified timeout
```
cgillum, line 64:
```suggestion
     * Sets a timeout duration for the purge operation. Setting to {@code null} will reset the timeout to be the default value.
```

cgillum (approved):
:shipit:

PR #106, opened 2023-01-23
Implement set timer by time stamp using Java.time.ZonedDateTime

### LINKED ISSUE #103: Ability to specify timestamp while creating durable timers
In the java SDK, We can only specify a Delay after which the timer will expire. Can we also start supporting timestamp-based durable timers in the JAVA version?
Ref: https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-timers?tabs=java

kaibocai:

resolves #103

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/Helpers.java
+++ b/client/src/main/java/com/microsoft/durabletask/Helpers.java
@@ -5,6 +5,7 @@ package com.microsoft.durabletask;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.time.Duration;
+import java.time.Instant;
 
 final class Helpers {
     final static Duration maxDuration = Duration.ofSeconds(Long.MAX_VALUE, 999999999L);
@@ -29,6 +30,12 @@ final class Helpers {
     static void throwIfOrchestratorComplete(boolean isComplete) {
         if (isComplete) {
             throw new IllegalStateException("The orchestrator has already completed");
+        }
+    }
+
+    static void throwIfFireAtBeforeCurrentInstant(Instant fireAt, Instant currentInstant) {
+        if (fireAt.isBefore(currentInstant)) {
+            throw new IllegalStateException("Timer cannot be earlier than current time stamp. ");
         }
     }
 
--- a/client/src/main/java/com/microsoft/durabletask/Helpers.java
+++ b/client/src/main/java/com/microsoft/durabletask/Helpers.java
@@ -5,7 +5,6 @@ package com.microsoft.durabletask;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.time.Duration;
-import java.time.Instant;
 
 final class Helpers {
     final static Duration maxDuration = Duration.ofSeconds(Long.MAX_VALUE, 999999999L);
@@ -30,12 +29,6 @@ final class Helpers {
     static void throwIfOrchestratorComplete(boolean isComplete) {
         if (isComplete) {
             throw new IllegalStateException("The orchestrator has already completed");
-        }
-    }
-
-    static void throwIfFireAtBeforeCurrentInstant(Instant fireAt, Instant currentInstant) {
-        if (fireAt.isBefore(currentInstant)) {
-            throw new IllegalStateException("Timer cannot be earlier than current time stamp. ");
         }
     }
 
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
@@ -167,7 +167,7 @@ public interface TaskOrchestrationContext {
      * @return a new {@code Task} that completes after the specified delay
      */
     default Task<Void> createTimer(ZonedDateTime zonedDateTime) {
-        throw new UnsupportedOperationException("This method is not implemented, please use Class has implementation for this method.");
+        throw new UnsupportedOperationException("This method is not implemented. Please use an orchestration context class that has an implementation for this method.");
     }
 
     /**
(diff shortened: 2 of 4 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
```diff
@@ -155,6 +156,18 @@ default Task<Task<?>> anyOf(Task<?>... tasks) {
...
+     * Creates a durable timer that expires after the specified timestamp with specific zone.
+     * <p>
+     * Specifying a long delay (for example, a delay of a few days or more) may result in the creation of multiple,
+     * internally-managed durable timers. The orchestration code doesn't need to be aware of this behavior. However,
+     * it may be visible in framework logs and the stored history state.
+     *
+     * @param zonedDateTime timestamp with specific zone when the timer should expire
+     * @return a new {@code Task} that completes after the specified delay
+     */
+    Task<Void> createTimer(ZonedDateTime zonedDateTime);
```
cgillum, line 169: Is it a breaking change to add a new interface method without a default implementation? Now that we’re stable, we should try not to introduce breaking changes. Not a big deal for this case since I don’t expect any 3rd party implementations to exist yet, but something to consider.

client/src/main/java/com/microsoft/durabletask/Helpers.java
```diff
@@ -32,6 +33,12 @@ static void throwIfOrchestratorComplete(boolean isComplete) {
         }
     }
 
+    static void throwIfFireAtBeforeCurrentInstant(Instant fireAt, Instant currentInstant) {
+        if (fireAt.isBefore(currentInstant)) {
+            throw new IllegalStateException("Timer cannot be earlier than current time stamp. ");
```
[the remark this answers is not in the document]
kaibocai: I understand that the timer can only set a future timestamp right, if the time stamp to fire at is earlier than current timestamp, it should not be possible and throw an exception. For example customer can not set up a timer at 2020-01-01 right. Or I am understanding it wrongly.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
cgillum: It’sa breaking change for anyone that creates their own implementation of this interface. For example, if someone writes unit tests for their orchestration logic by implementing a mock task orchestration context, this change may break their test code. Similarly, if a user creates their own higher level abstraction around orchestrations and implements this interface, that abstraction will break.

kaibocai: I add an default implmentation by throwing an `UnsupportedOperationException` to avoid any breaking change.

client/src/main/java/com/microsoft/durabletask/Helpers.java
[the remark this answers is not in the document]
kaibocai: remove the logic to check if timer is set in the past.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
```diff
@@ -155,6 +156,20 @@ default Task<Task<?>> anyOf(Task<?>... tasks) {
...
+     * <p>
+     * Specifying a long delay (for example, a delay of a few days or more) may result in the creation of multiple,
+     * internally-managed durable timers. The orchestration code doesn't need to be aware of this behavior. However,
+     * it may be visible in framework logs and the stored history state.
+     *
+     * @param zonedDateTime timestamp with specific zone when the timer should expire
+     * @return a new {@code Task} that completes after the specified delay
+     */
+    default Task<Void> createTimer(ZonedDateTime zonedDateTime) {
+        throw new UnsupportedOperationException("This method is not implemented, please use Class has implementation for this method.");
```
cgillum, line 170:
```suggestion
        throw new UnsupportedOperationException("This method is not implemented. Please use an orchestration context class that has an implementation for this method.");
```

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -530,6 +531,20 @@ public Task<Void> createTimer(Duration duration) {
...
+            return createInstantTimer(id, fireAt);
+        }
+
+        @Override
+        public Task<Void> createTimer(ZonedDateTime zonedDateTime) {
+            Helpers.throwIfOrchestratorComplete(this.isComplete);
+            Helpers.throwIfArgumentNull(zonedDateTime, "zonedDateTime");
+
+            int id = this.sequenceNumber++;
+            Instant fireAt = zonedDateTime.toInstant();
```
cgillum, line 543: The timestamps we use must always be in UTC. Do you know if this `zonedDateTime.toInstant()` produces an `Instant` in UTC (or if not, do we know if the call to `CreateTimerAction.newBuilder().setFireAt(ts)` converts the time into UTC)?

kaibocai: `Instant` are not Timezone aware. So whatever time zone we have in `zonedDateTime` when transfer to `Instant` it's always UTC. 

> An Instant is a moment on the timeline in UTC, a count of nanoseconds since the epoch of the first moment of 1970 UTC

Local test example:

```java
    public static void main(String[] args) {
        System.out.println(ZonedDateTime.now(ZoneId.of("Australia/Darwin")).toInstant());
        System.out.println(ZonedDateTime.now(ZoneId.of("Europe/Paris")).toInstant());
    }
```

image

cgillum: Awesome - thanks for clarifying and providing this example!

cgillum (approved):

PR #116, opened 2023-02-27
Add end to end test suits. 

kaibocai:

Add end to end test suits. 
### Pull request checklist

### Additional information

Additional PR information


samples-azure-functions/e2e-test.ps1
```diff
@@ -41,55 +39,4 @@ if ($sleep -gt  0) {
 }
 
 # Check to see what containers are running
-docker ps
-
-try {
```
cgillum, line 46: It looks like we're no longer running an end-to-end test with this script. Should we rename it to something else?

samples-azure-functions/src/main/java/com/functions/SuspendResumeFunctions.java
```diff
@@ -0,0 +1,65 @@
...
+import com.microsoft.durabletask.azurefunctions.DurableOrchestrationTrigger;
+
+import java.util.Optional;
+
+public class SuspendResumeFunctions {
+    /**
+     * This HTTP-triggered function starts the orchestration.
+     */
+    @FunctionName("SuspendResumeOrchestration")
+    public HttpResponseMessage suspendResumeOrchestration(
```
cgillum, line 24: How is this test getting invoked?

samples-azure-functions/src/main/java/com/functions/SuspendResumeFunctions.java
```diff
@@ -0,0 +1,65 @@
...
+     * This is the orchestrator function, which can schedule activity functions, create durable timers,
+     * or wait for external events in a way that's completely fault-tolerant.
+     */
+    @FunctionName("SuspendResumeCities")
+    public String Cities(
+            @DurableOrchestrationTrigger(name = "ctx") TaskOrchestrationContext ctx) throws InterruptedException {
+        String result = "";
+        result += ctx.callActivity("Capitalize", "Tokyo", String.class).await() + ", ";
+        result += ctx.callActivity("Capitalize", "London", String.class).await() + ", ";
+        Thread.sleep(5000);
```
cgillum, line 49: Should we use `ctx.createTimer()` instead of `Thread.sleep()` here? Technically `Thread.sleep()` is considered illegal.

samples-azure-functions/src/main/java/com/functions/SuspendResumeFunctions.java
kaibocai: Yes, this is not tested yet. I will remove this one for now, cause seems it hasn't be available in current function runtime. Once this feature is available, will add end-to-end tests for it. Thanks.

cgillum (approved):

PR #122, opened 2023-03-06
Break long timers into smaller timers

### LINKED ISSUE #114: [Eternal Orchestrator] Timers more than 7days giving error
I have implemented a scheduler out of an eternal orchestrator. This scheduler evaluates a cron expression and calculates the next execution time based on it. This eternal orchestrator now needs to sleep till this next execution time. 
I was reading the documentation on timers in durable framework, and according to [this](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-timers?tabs=java#timer-limitations) there is no limitation on timers for java, however when I tried it I am getting this error : 
` Microsoft.WindowsAzure.Storage: The argument 'initialVisibilityDelay' is larger than maximum of '7.00:00:00' (Parameter 'initialVisibilityDelay').
`

Can you help me understand if this is a bug or am I doing something wrong. Here the delay I am generating is of 44 days.

kamperiadis:

Resolves #114 
Currently, we do not support timers longer than 7 days. However, the changes in this PR work around that by breaking up a single long timer into several shorter ones.

### Pull request checklist

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
@@ -11,6 +11,7 @@ import com.microsoft.durabletask.implementation.protobuf.TaskHubSidecarServiceGr
 
 import io.grpc.*;
 
+import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
@@ -28,6 +29,8 @@ public final class DurableTaskGrpcWorker implements AutoCloseable {
 
     private final ManagedChannel managedSidecarChannel;
     private final DataConverter dataConverter;
+    private static final Duration DEFAULT_MAXIMUM_TIMER_INTERVAL = Duration.ofDays(3);
+    private final Duration maximumTimerInterval;
 
     private final TaskHubSidecarServiceBlockingStub sidecarClient;
 
@@ -57,6 +60,7 @@ public final class DurableTaskGrpcWorker implements AutoCloseable {
 
         this.sidecarClient = TaskHubSidecarServiceGrpc.newBlockingStub(sidecarGrpcChannel);
         this.dataConverter = builder.dataConverter != null ? builder.dataConverter : new JacksonDataConverter();
+        this.maximumTimerInterval = builder.maximumTimerInterval != null ? builder.maximumTimerInterval : DEFAULT_MAXIMUM_TIMER_INTERVAL;
     }
 
     /**
@@ -108,6 +112,7 @@ public final class DurableTaskGrpcWorker implements AutoCloseable {
         TaskOrchestrationExecutor taskOrchestrationExecutor = new TaskOrchestrationExecutor(
                 this.orchestrationFactories,
                 this.dataConverter,
+                this.maximumTimerInterval,
                 logger);
         TaskActivityExecutor taskActivityExecutor = new TaskActivityExecutor(
                 this.activityFactories,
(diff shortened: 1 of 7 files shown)


client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -28,6 +29,8 @@ public final class DurableTaskGrpcWorker implements AutoCloseable {
 
     private final ManagedChannel managedSidecarChannel;
     private final DataConverter dataConverter;
+    private static final Duration DEFAULT_MAXIMUM_TIMER_INTERVAL = Duration.ofDays(3);
```
cgillum, line 32: nit: move this up so that all the `private static final` fields can be grouped together.

client/src/test/java/com/microsoft/durabletask/IntegrationTests.java
```diff
@@ -86,6 +86,30 @@ void singleTimer() throws IOException, TimeoutException {
         }
     }
 
+    @Test
+    void longTimer() throws TimeoutException {
+        final String orchestratorName = "LongTimer";
+        final Duration delay = Duration.ofSeconds(7);
+        DurableTaskGrpcWorker worker = this.createWorkerBuilder()
+                .addOrchestrator(orchestratorName, ctx -> ctx.createTimer(delay).await())
```
cgillum, line 94: One issue with this test is that we don't know if the one timer was actually broken into three timers or not. One way you could verify this is by creating a counter variable outside of the orchestrator lambda function and incrementing it each time the orchestrator lambda function executes. Since we expect the 7 seconds to be distributed across 3 timers (3s + 3s + 1s = 7s), I think we can add an assert at the end of this test to test whether the counter was set to `4` (first invocation + a replay invocation for each internal timer).

cgillum (commented):
This looks great! Just a couple small suggestions from me.

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorkerBuilder.java
```diff
@@ -99,6 +101,17 @@ public DurableTaskGrpcWorkerBuilder dataConverter(DataConverter dataConverter) {
         return this;
     }
 
+    /**
+     * Sets the maximum timer interval. If not specified, the default maximum timer interval duration will be used.
```
kaibocai, line 105: we can mention the current default duration is 3 days.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -547,9 +551,14 @@ public Task<Void> createTimer(Duration duration) {
...
 
-            int id = this.sequenceNumber++;
-            Instant fireAt = this.currentInstant.plus(duration);
-            return createInstantTimer(id, fireAt);
+            Instant finalFireAt = this.currentInstant.plus(duration);
+            Duration remainingTime = Duration.between(finalFireAt, this.currentInstant);
+            while (remainingTime.compareTo(this.maximumTimerInterval) > 0) {
+                Instant nextFireAt = this.currentInstant.plus(this.maximumTimerInterval);
+                createInstantTimer(this.sequenceNumber++, nextFireAt).await();
+                remainingTime = Duration.between(finalFireAt, this.currentInstant);
```
kaibocai, line 559: Similar logics should also be added to method `createTimer(ZoneDateTime)`

kaibocai (commented):
LGTM, add 2 minnor comments.

kaibocai (approved):

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -543,23 +547,31 @@ private void handleExecutionResumed(HistoryEvent historyEvent) {
             }
         }
 
+        private Task<Void> createTimer(Instant finalFireAt) {
```
cgillum, line 550: Minor nit: To make the code easier to follow from top-to-bottom, consider moving this method below the other two `createTimer` method overloads. It's not important to do this, but it's a common practice I like to follow so that anyone who reads the code doesn't have to scroll up and down so much to understand the flow of creating timers, etc.

cgillum (approved):
LGTM (Looks Good To Me)!

Just one minor suggestion which you can choose to incorporate or skip.

kanupriya15025:
@kamperiadis Is this released in the SDK?

PR #123, opened 2023-03-07
Add rewind client API

### LINKED ISSUE #65: Add rewind client API
See https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management?tabs=csharp#rewind-instances-preview.

Once we have this API, the link documentation should be updated with a sample (some sample code actually already exists but is commented out).

shreyas-gopalakrishna:

resolves #65

### Pull request checklist

### Additional information

Additional PR information


samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
```diff
@@ -36,4 +38,36 @@ public void basicChain() throws InterruptedException {
         }
         assertEquals("Completed", runTimeStatus);
     }
+
+    @Order(2)
+    @Test
+    public void approvalWorkFlow() throws InterruptedException {
```
shreyas-gopalakrishna, line 44: @cgillum Checking if this test case is good for verifying rewind instance.

samples-azure-functions/extensions.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <WarningsAsErrors></WarningsAsErrors>
+    <DefaultItemExcludes>**</DefaultItemExcludes>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.9.2" />
```
shreyas-gopalakrishna, line 8: Using this config is currently resulting in the below exception. Should we wait for next bundle release?
MicrosoftTeams-image

cgillum: @davidmrdavid how close are we to completing the v2.9.3 extension release?

samples-azure-functions/src/main/java/com/functions/RewindInstance.java
```diff
@@ -0,0 +1,91 @@
...
+        String instanceId = request.getQueryParameters().getOrDefault("instanceId", "");
+        String reason = "Orchestrator failed and needs to be revived.";
+
+        DurableTaskClient client = durableContext.getClient();
+
+        try {
+            client.rewindInstance(instanceId, reason);
+            return "Failed orchestration instance is revived.";
+        }catch (Exception e){
+            return "Exception when rewinding orchestration instance";
```
cgillum, line 88: Why not just let the exception escape? This isn't very useful as-is since we don't even log anything about the exception.

samples-azure-functions/src/main/java/com/functions/RewindInstance.java
```diff
@@ -0,0 +1,91 @@
...
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            final ExecutionContext context) {
+        String instanceId = request.getQueryParameters().getOrDefault("instanceId", "");
+        String reason = "Orchestrator failed and needs to be revived.";
+
+        DurableTaskClient client = durableContext.getClient();
+
+        try {
+            client.rewindInstance(instanceId, reason);
+            return "Failed orchestration instance is revived.";
```
cgillum, line 86: This isn't really accurate. The orchestration gets revived asynchronously, so it may be another second (or longer) until it actually revives. It's more correct to say that we _scheduled_ the rewind.

samples-azure-functions/src/main/java/com/functions/RewindInstance.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * This is the activity function that fails the first try and is then revived.
+     */
+    @FunctionName("RequestSecondaryApproval")
+    public int requestSecondaryApproval(
+            @DurableActivityTrigger(name = "name") int number,
+            final ExecutionContext context) throws InterruptedException {
+        System.out.println("Test RequestSecondaryApproval " + approvalFlag);
```
cgillum, line 67: Should we use a logger instead? That way the sample can be run in Azure and the output can be seen in Application Insights.

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
cgillum, line 44: Let's rename this to make it clear that we're testing rewind functionality. How about `testRewindHttpAPI()`?

cgillum: The scenario is good, but let's combine these two tests into one. This first test, which just waits for the orchestration to fail, is not meaningful by itself.

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
```diff
@@ -36,4 +38,36 @@ public void basicChain() throws InterruptedException {
         }
         assertEquals("Completed", runTimeStatus);
     }
+
+    @Order(2)
+    @Test
+    public void approvalWorkFlow() throws InterruptedException {
+        Response response = post(approvalWorkFlow);
+        rewindTestJsonPath = response.jsonPath();
```
cgillum, line 46: I'm confused by what this test is doing. What is a `jsonPath`? Can you provide an example?

samples-azure-functions/src/main/java/com/functions/RewindInstance.java
```diff
@@ -0,0 +1,91 @@
...
+        return number / approvalFlag++;
+    }
+
+    /**
+     * This HTTP-triggered function rewinds the orchestration using instanceId.
+     */
+    @FunctionName("RewindInstance")
+    public String rewindInstance(
```
cgillum, line 75: Are we invoking this API as part of the end-to-end test? I see in the test logic that we're invoking the built-in rewind HTTP API but I couldn't tell if we're calling this API (which invokes the client API).

cgillum (changes requested):

samples-azure-functions/src/main/java/com/functions/RewindInstance.java
shreyas-gopalakrishna: This log was for my local debugging. I have removed it

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
shreyas-gopalakrishna: jsonPath here holds the json response when the orchestration is started. It is similar to HttpCreateCheckStatusResponse

```json
{
  "id": "d05dd7ea",
  "purgeHistoryDeleteUri": "",
  "sendEventPostUri": "",
  "statusQueryGetUri": "",
  "terminatePostUri": "",
  "rewindPostUri": ""
}
```

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
```diff
@@ -36,4 +38,37 @@ public void basicChain() throws InterruptedException {
         }
         assertEquals("Completed", runTimeStatus);
     }
+
+    @Order(2)
+    @Test
+    public void testRewindInstanceAPI() throws InterruptedException {
+        Response response = post(approvalWorkFlow);
+        JsonPath rewindTestJsonPath = response.jsonPath();
```
cgillum, line 46: I'm not sure where `Response` or `JsonPath` are defined, but [JsonPath](https://goessner.net/articles/JsonPath/) means something very specific in the industry that's different from what we have here.

I don't think it's worth changing class names in this PR, but can we change `rewindTestJsonPath` to something that doesn't include the word "path"? It's quite confusing for me. How about `startOrchestrationResponseJson`?

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
```diff
@@ -36,4 +38,37 @@ public void basicChain() throws InterruptedException {
...
+        Response response = post(approvalWorkFlow);
+        JsonPath rewindTestJsonPath = response.jsonPath();
+
+        // Wait for the ApprovalWorkflowOrchestration to fail
+        Thread.sleep(3000);
+
+        String instanceId = rewindTestJsonPath.get("id");
+        String statusQueryGetUri = rewindTestJsonPath.get("statusQueryGetUri");
+        Response statusResponse = get(statusQueryGetUri);
+        String runTimeStatus = statusResponse.jsonPath().get("runtimeStatus");
```
cgillum, line 54: nit: "runtime" is one word.
```suggestion
        String runtimeStatus = statusResponse.jsonPath().get("runtimeStatus");

```

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
```diff
@@ -14,6 +14,8 @@
 public class EndToEndTests {
     private static final String hostHealthPingPath = "/admin/host/ping";
     private static final String startOrchestrationPath = "/api/StartOrchestration";
+    private static final String approvalWorkFlow = "/api/ApprovalWorkflowOrchestration";
```
cgillum, line 17: nit: "workflow" is one word. Also, it's good practice to be more clear about what a variable represents. In this case it doesn't represent a "workflow" but a URL for starting a workflow, so a better variable name would be `startApprovalWorkflowUrl`
```suggestion
    private static final String startApprovalWorkflowUrl = "/api/ApprovalWorkflowOrchestration";

```

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
```diff
@@ -14,6 +14,8 @@
 public class EndToEndTests {
     private static final String hostHealthPingPath = "/admin/host/ping";
     private static final String startOrchestrationPath = "/api/StartOrchestration";
+    private static final String approvalWorkFlow = "/api/ApprovalWorkflowOrchestration";
+    private static final String rewindInstance = "/api/RewindInstance";
```
cgillum, line 18: Same comment about variable naming:
```suggestion
    private static final String rewindInstanceFunctionUrl = "/api/RewindInstance";

```

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
```diff
@@ -36,4 +38,37 @@ public void basicChain() throws InterruptedException {
...
+
+        String instanceId = rewindTestJsonPath.get("id");
+        String statusQueryGetUri = rewindTestJsonPath.get("statusQueryGetUri");
+        Response statusResponse = get(statusQueryGetUri);
+        String runTimeStatus = statusResponse.jsonPath().get("runtimeStatus");
+        assertEquals("Failed", runTimeStatus);
+
+        // Rewind the instance
+        String rewindPostUri = rewindInstance + "?instanceId=" + instanceId;
+        response = post(rewindPostUri);
```
cgillum, line 59: It's good that we're invoking the Java API for rewinding instances, but I think it's also important for us to have test coverage for the built-in HTTP API case as well (like you had in the previous iteration). I suggest we have two end-to-end tests using the same test functions - one that rewinds using the built-in HTTP API and one that rewinds using the Java API.

cgillum (changes requested):
Added more comments.

Moving forward, please be especially mindful about variable naming. Poor variable names can make it a lot harder to correctly understand and review code. It can also be a distraction. Good variable naming, on the other hand, can make everything much easier to understand quickly and makes the overall code feel more professional and polished.

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
shreyas-gopalakrishna: Thank you for the feedback Chris. I will improve on following better naming convention.

samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
shreyas-gopalakrishna: Updated to have tests for both the scenarios

samples-azure-functions/extensions.csproj
cgillum: Looks like the v2.9.3 extension release is in the validation stage right now.

cgillum (dismissed):
This looks great - thanks!

samples-azure-functions/extensions.csproj
kaibocai: @shreyas-gopalakrishna , Can we try again without this .csproj file, the v2.9.3 should be included in the extension bundle by now. Thanks.

shreyas-gopalakrishna (commented):
image

image

Facing the above proto issue when running rewind instance on ExtensionVersion: 2.10.0
cc @cgillum @kaibocai if this has been encontered before

PR #131, opened 2023-04-21
Set up CI with Azure Pipelines

kaibocai:


cgillum (commented):
What is this pipeline for? I see it does a build, but is there a plan to include a release step?

kaibocai:
> What is this pipeline for? I see it does a build, but is there a plan to include a release step?

ESRP extension is now only integrated on azure devops, so we need to set an azure devops pipeline to use ESRP for our release.

Issue #134, opened 2023-05-13
Deserialize problem on LocalDate


nicktombeur:
When returning an object containing a LocalDate, I get the following error.

```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
 at [Source: (String)"{"date":{"year":2023,"month":5,"day":13},"value":"foo"}"; line: 1, column: 1] 
...
 at com.microsoft.durabletask.JacksonDataConverter.deserialize(JacksonDataConverter.java:40)
```

I believe this date format is coming from protobuf?
I can solve this problem by writing my own custom Jackson deserializer but shouldn't this be handled for us?
Maybe it's possible to return` { "date": "2023-05-13" }` instead? This way  I can add `jackson-datatype-jsr310` to the class path and `com.microsoft.durabletask.JacksonDataConverter` will pick it up. 

Example code:

```
public class ExampleFunction {

    @FunctionName("StartOrchestration")
    public HttpResponseMessage startOrchestration(
            @HttpTrigger(name = "req",
                    methods = {HttpMethod.GET, HttpMethod.POST},
                    authLevel = AuthorizationLevel.ANONYMOUS) final HttpRequestMessage<Optional<String>> request,
            @DurableClientInput(name = "durableContext") final DurableClientContext durableContext,
            final ExecutionContext context) {
        context.getLogger().info("Java HTTP trigger processed a request");

        final DurableTaskClient client = durableContext.getClient();
        final String instanceId = client.scheduleNewOrchestrationInstance("ExampleProcess");
        return durableContext.createCheckStatusResponse(request, instanceId);
    }

    @FunctionName("ExampleProcess")
    public String exampleOrchestrator(
            @DurableOrchestrationTrigger(name = "taskOrchestrationContext") final TaskOrchestrationContext context,
            final ExecutionContext functionContext) {
        return context.callActivity("ToLower", "Foo", String.class).await() +
                    " " +
                    context.callActivity("ToLower", "Bar", String.class).await();
    }

    @FunctionName("ToLower")
    public ExampleResponse toLower(@DurableActivityTrigger(name = "value") final String value, final ExecutionContext context) {
        return new ExampleResponse(LocalDate.now(), value.toLowerCase());
    }
}

public class ExampleResponse {
    private LocalDate date;
    private String value;

   ...
}
```

kaibocai:
[...]

PR #145, opened 2023-06-23
Provide a way for customer to inject their own data converter. 

kaibocai:

Try to improve https://github.com/microsoft/durabletask-java/issues/134. This PR provides the customer a way to provide their own data converter through java SPI. To have it work on java11 and above, the dependent PR https://github.com/Azure/azure-functions-java-worker/pull/717 has to be released in java worker. 

~~Have to do more tests on this PR. Will try to provide more details once the PR is ready.~~
Tested on local worked as expected. 

Right now for the issue mentioned in https://github.com/microsoft/durabletask-java/issues/134, using my customized data converter based on gson works fine

The end2end test failed because the java worker is not released yet, this feature shall wait for the PR at https://github.com/Azure/azure-functions-java-worker/pull/717/files to be released first. 
Once the image here https://github.com/microsoft/durabletask-java/blob/57133bdee6717b11186cc84d7cf849a44d48d6a5/samples-azure-functions/Dockerfile#L1 include above java worker PR then the end2end test should pass. 

image

cc @nicktombeur

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
+++ b/samples-azure-functions/src/test/java/com/functions/EndToEndTests.java
@@ -15,6 +15,18 @@ import static org.junit.jupiter.api.Assertions.*;
 @Tag("e2e")
 public class EndToEndTests {
 
+    private String waitForCompletion(String statusQueryGetUri) throws InterruptedException {
+        String runTimeStatus = null;
+        for (int i = 0; i < 15; i++) {
+            Response statusResponse = get(statusQueryGetUri);
+            runTimeStatus = statusResponse.jsonPath().get("runtimeStatus");
+            if (!"Completed".equals(runTimeStatus)) {
+                Thread.sleep(1000);
+            } else break;
+        }
+        return runTimeStatus;
+    }
+
     @Order(1)
     @Test
     public void setupHost() {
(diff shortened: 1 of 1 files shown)


azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
```diff
@@ -39,12 +49,29 @@ public void invoke(MiddlewareContext context, MiddlewareChain chain) throws Exce
...
+            synchronized (dataConverterLock) {
+                if (!oneTimeLogicExecuted.get()) {
+                    Iterator<DataConverter> iterator = ServiceLoader.load(DataConverter.class).iterator();
+                    if (iterator.hasNext()) {
+                        this.dataConverter = iterator.next();
+                        if (iterator.hasNext()) {
+                            throw new IllegalStateException("Multiple implementations of DataConverter found on the classpath.");
+                        }
+                    }
+                    oneTimeLogicExecuted.compareAndSet(false,true);
```
cgillum, line 72: Just curious why you can't use a regular `Boolean` here? Since we're already in a `synchronized` block, I wouldn't expect that you'd need to use any additional synchronization primitives.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
```diff
@@ -39,12 +49,29 @@ public void invoke(MiddlewareContext context, MiddlewareChain chain) throws Exce
...
                 // invoke method. Thus get the cause to check if it's OrchestratorBlockedEvent.
                 Throwable cause = e.getCause();
-                if (cause instanceof OrchestratorBlockedException){
+                if (cause instanceof OrchestratorBlockedException) {
                     throw (OrchestratorBlockedException) cause;
                 }
                 throw new RuntimeException("Unexpected failure in the task execution", e);
             }
-        });
+        }, dataConverter);
```
cgillum, line 57: To make it clear this is a member variable and not a local variable:
```suggestion
        }, this.dataConverter);

```

cgillum (approved):
Nice!

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
kaibocai: If I use a regular `Boolean` in java, I need to declare it with `volatile` which avoids using thread cache, that means that when a thread modifies the value of this `Boolean`, all other threads will see the updated value when they access it in Java. So I am thinking of just using AtomicBoolean which provides both visibility and atomicity at the same time.

ScottHamper:
Hey All,

Any chance this feature can get merged soon? I've been experimenting with Azure Durable Functions using Kotlin, and desperately wish I could use kotlinx serialization instead of Jackson.

Issue #118, opened 2023-02-28
Bug in functioning of continueAsNew() function in eternal orchestration


kanupriya15025:
According to the documentation here (https://github.com/microsoft/durabletask-java/blob/6d6dff9bbbe145f561cde05ff7ab065530279073/README.md#eternal-monitoring-orchestration) , if continueAsNew is called, the flow should break and the a new execution of the eternal orchestrator should be called. However, that's not what I am seeing here in my code : 

            if (winnerEvent == patchEvent) {
                jsonBody = patchEvent.await();
                if (!ctx.getIsReplaying()) {
                    LOGGER.info("Payload received for PATCH call: {}", jsonBody);
                    ctx.continueAsNew(jsonBody);
                }
            }

            LOGGER.info("The control passed here.");

In the above code my expectation is that as soon as ctx.continueAsNew() is called, the flow should not be passed to the second log line. But this is what is what I see :

            Payload received for PATCH call: <>
            The control passed here.

My expectation is that if winnerEvent is PatchEvent, I should only see log line "Payload received for PATCH call:" and not "The control passed here"

Is this expectation correct?

davidmrdavid:
Hi @kanupriya15025:

Given that `continueAsNew` is not an "awaited" API, we _normally_ cannot stop the function from continuing to execute any lines of code directly below it. However, on the next opportunity where execution control is yielded back to the SDK (i.e an API is "awaited" or the code returns) we'll execute the continueAsNew command.

I say that _normally_ we cannot stop the function from continuing to execute because I'm not aware if the Durable Functions Java SDK design may have something special that allows it to break the control flow immediately when continueAsNew is called. @kaibocai - let's sync about this offline.

cgillum:
@kanupriya15025 for now, the correct thing to do is to `return` immediately after your call to `continueAsNew()`.

ChrisRomp:
FWIW, I'm still getting the exception even after a `return` statement from a `void` method.

Code:
```java
@FunctionName("Periodic_Cleanup_Loop")
public void periodicCleanupLoop(
        @DurableOrchestrationTrigger(name = "ctx") TaskOrchestrationContext ctx) {
    
    ctx.createTimer(Duration.ofHours(1)).await();
    ctx.continueAsNew(null);
    return;
}
```

API call:
```http
POST http://localhost:7071/runtime/webhooks/durabletask/orchestrators/{{functionName}}/{{instanceId}}?code={{systemKey}} HTTP/1.1
[...]

PR #150, opened 2023-07-05
Improve continueAsNew

kaibocai:

Try to improve the issue mentioned here https://github.com/microsoft/durabletask-java/issues/118#issuecomment-1549064214

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/Task.java
+++ b/client/src/main/java/com/microsoft/durabletask/Task.java
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 package com.microsoft.durabletask;
 
+import com.microsoft.durabletask.interruption.OrchestratorBlockedException;
+
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 import java.util.function.Function;
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
@@ -296,7 +296,7 @@ final class TaskOrchestrationExecutor {
             this.preserveUnprocessedEvents = preserveUnprocessedEvents;
 
             // The ContinueAsNewInterruption exception allows the orchestration to complete immediately and return back
-            // to azure functions durable extension/sidecar.
+            // to the sidecar.
             // We can send the current set of actions back to the worker and wait for new events to come in.
             // This is *not* an exception - it's a normal part of orchestrator control flow.
             throw new ContinueAsNewInterruption(
@@ -751,9 +751,7 @@ final class TaskOrchestrationExecutor {
             }
 
             if (this.continuedAsNew && this.preserveUnprocessedEvents) {
-                for (HistoryEvent e : this.unprocessedEvents) {
-                    builder.addCarryoverEvents(e);
-                }
+                addCarryoverEvents(builder);
             }
 
             if (!this.isReplaying) {
@@ -766,6 +764,21 @@ final class TaskOrchestrationExecutor {
                     .build();
             this.pendingActions.put(id, action);
             this.isComplete = true;
+        }
+
+        private void addCarryoverEvents(CompleteOrchestrationAction.Builder builder) {
+            // Add historyEvent in the buffer
+            for (HistoryEvent e : this.unprocessedEvents) {
+                builder.addCarryoverEvents(e);
+            }
+            // Add historyEvent in the new event list that haven't been added to the buffer.
+            // We don't check the event in the pass event list to avoid duplicated events.
+            List<HistoryEvent> newEvents = this.historyEventPlayer.getNewEvents();
+            for (HistoryEvent e : newEvents) {
+                if (e.getEventTypeCase() == HistoryEvent.EventTypeCase.EVENTRAISED) {
+                    builder.addCarryoverEvents(e);
+                }
+            }
         }
         
         private boolean waitingForEvents() {
@@ -915,6 +928,10 @@ final class TaskOrchestrationExecutor {
                 HistoryEvent next = this.currentHistoryList.get(this.currentHistoryIndex++);
                 ContextImplTask.this.processEvent(next);
                 return true;
+            }
+
+            List<HistoryEvent> getNewEvents() {
+                return newEvents;
             }
         }
 
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
@@ -295,6 +295,10 @@ final class TaskOrchestrationExecutor {
             this.continuedAsNewInput = input;
             this.preserveUnprocessedEvents = preserveUnprocessedEvents;
 
+            // If there are events haven't been processed, we make sure to finish processing them before the interruption,
+            // so no event is dropped.
+            while(processNextEvent()) {}
+
             // The ContinueAsNewInterruption exception allows the orchestration to complete immediately and return back
             // to azure functions durable extension/sidecar.
             // We can send the current set of actions back to the worker and wait for new events to come in.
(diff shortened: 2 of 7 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -289,6 +294,17 @@ public void continueAsNew(Object input, boolean preserveUnprocessedEvents) {
             this.continuedAsNew = true;
             this.continuedAsNewInput = input;
             this.preserveUnprocessedEvents = preserveUnprocessedEvents;
+
+            // If there are events haven't been processed, we make sure to finish processing them before the interruption,
+            // so no event is dropped.
+            while(processNextEvent()) {}
+
+            // The ContinueAsNewInterruption exception allows the orchestration to complete immediately and return back
+            // to azure functions durable extension/sidecar.
```
cgillum, line 303: I recommend just calling this a sidecar. Whether it's Azure Functions or Dapr Workflow or the .NET sidecar, it can still be considered a "sidecar". I worry that talking about Azure Functions in a non-Azure Functions project could confuse people.
```suggestion
            // to the sidecar.
```

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -289,6 +294,17 @@ public void continueAsNew(Object input, boolean preserveUnprocessedEvents) {
             this.continuedAsNew = true;
             this.continuedAsNewInput = input;
             this.preserveUnprocessedEvents = preserveUnprocessedEvents;
+
+            // If there are events haven't been processed, we make sure to finish processing them before the interruption,
+            // so no event is dropped.
+            while(processNextEvent()) {}
```
cgillum, line 300: I don't quite understand what the implications of processing all the events like this is. Won't this continue to run the user code, which is what we're trying to avoid?

Why do we need to finish processing events?

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -1090,6 +1106,10 @@ public V await() {
                         try {
                             return this.future.get();
                         } catch (ExecutionException e) {
+                            // rethrow if it's ContinueAsNewInterruption
+                            if (e.getCause() instanceof ContinueAsNewInterruption) {
+                                throw (ContinueAsNewInterruption) e.getCause();
```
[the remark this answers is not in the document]
kaibocai: I think that's not how the catch block works. In Java, catch blocks are designed to handle exceptions that occur within their corresponding try block.

For example:
image

To catch that exception I need to wrap the entire code with another `try-catch` block like
image

client/src/main/java/com/microsoft/durabletask/interruption/OrchestratorBlockedException.java
```diff
@@ -12,7 +14,7 @@
  * state and scheduling new tasks, resulting in the orchestration getting stuck.
  */
 public final class OrchestratorBlockedException extends RuntimeException {
-    OrchestratorBlockedException(String message) {
+    public OrchestratorBlockedException(String message) {
```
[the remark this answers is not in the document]
kaibocai: I refactor it to a subpackage, which requires it to be `public` to be allowed to use in files not in the same subpackage. 
image

With all files put under the same packages is not quite usual in Java, having sub-packages for them can clarify the project.
But I am open to changing it back for better privacy of those classes.  

In Java the leading convention for this is to put them under a package named like
`com.microsoft.durabletask.internal.interruption`
(has `internal`, so users know they shouldn't use classes under this package)

and with a description at the head of those class like
```java
/**
 * This interface is implemented by middlewares to include middleware core logics.
 *
 * <p>This class is internal and is hence not for public use at this time. Its APIs are unstable and can change
 * at any time.
 */
```

client/src/main/java/com/microsoft/durabletask/Task.java
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 package com.microsoft.durabletask;
 
+import com.microsoft.durabletask.interruption.OrchestratorBlockedException;
```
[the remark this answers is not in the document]
kaibocai: It is used in the javadoc of this class at https://github.com/microsoft/durabletask-java/blob/1b2c05bd43d82934f8a9e83acb1944d5c6d70512/client/src/main/java/com/microsoft/durabletask/Task.java#L28

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
kaibocai: I updated 2 integration tests, adding a timer before sending out an event so that in each continueAsNew session there is only one event sent to it. Thus no event is dropped and causing the timeout for the test. 
However, is there a better way to do this?

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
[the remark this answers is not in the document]
cgillum: Got it - I got confused by the nesting and mistakenly thought that the catch statement below was an outer catch block.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
cgillum: Ah okay, now I understand the problem. You're losing external events that are sent after the first one because the continueAsNew logic, which gets triggered by the first external event, only saves external events that were buffered _before_ continueAsNew got called.

The test change you made simply hides the data loss problem, but customers will still run into it. We need to leave that test as-is and make sure that we do the work necessary to get it passing.

I think the way to fix this is to update the code [here](https://github.com/microsoft/durabletask-java/blob/3078c274270f7e2ec5132ee9f318d463a3e4d262/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java#L742-L744) to also search through the `newEvents` for any received external event messages and add them to `builder.addCarryoverEvents()`. You might have to access them from `historyEventPlayer` somehow.

kaibocai: Just want to confirm if we want to implement it in this way. Cause I think it's also reasonable to think that `continueAsNew` is not safe when waiting for external events in such a situation. As it will return the orchestration immediately and thus your events can be dropped.

cgillum: What you're saying is how `continueAsNew` was designed originally for .NET many years ago. However, demand from users to unblock certain scenarios is the reason why we added the [`preserveUnprocessedEvents`](https://github.com/microsoft/durabletask-java/blob/3078c274270f7e2ec5132ee9f318d463a3e4d262/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java#L311-L314) flag, which enables this event-preserving behavior.

kaibocai: Gotcha, will update the logic.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -767,6 +765,21 @@ private void completeInternal(
...
             this.isComplete = true;
         }
+
+        private void addCarryoverEvents(CompleteOrchestrationAction.Builder builder) {
+            // Add historyEvent in the buffer
+            for (HistoryEvent e : this.unprocessedEvents) {
+                builder.addCarryoverEvents(e);
+            }
+            // Add historyEvent in the new event list that haven't been added to the buffer.
+            // We don't check the event in the pass event list to avoid duplicated events.
```
cgillum, line 775: I'm worried that there may be duplication between `this.unprocessedEvents` and `getNewEvents()`. For example, if a batch of events comes in, the first one is an external event that gets buffered and added to `unprocessedEvents`, and then a second event (maybe a timer or a TASKCOMPLETED event) then causes the orchestration to invoke `continueAsNew(..., true)` it seems that buffered event will be added twice. Can we modify your solution to only consider the subset of `newEvents` which haven't been processed yet?

cgillum (approved):

Issue #149, opened 2023-07-05
Automatic retry not working with Fan-out Fan-in pattern


akshaykumars10:
I am trying to add automatic retry in my orchestrator function following this document: https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-error-handling?tabs=java#automatic-retry-on-failure

I am using the fan-out fan-in pattern in my orchestrator function. With retry logic, I can see my orchestrator function getting stuck in a running state in case of a failure. Do we need to do anything differently to have retry logic with this pattern?

This is what my orchestrator function looks like:

```
RetryPolicy policy = new RetryPolicy(maxRetry, firstRetryInterval);
TaskOptions options = new TaskOptions(policy);
String taskType = task.getType();
parallelTasks.add(ctx.callActivity(taskType, "input for activity function", options, String.class));
.
.
.
.
List<String> parallelResult = ctx.allOf(parallelTasks).await();
```

I have this config on my local setup:
Core Tools Version:       4.0.5198 Commit hash: N/A  (64-bit)
Function Runtime Version: 4.21.1.20667
durabletask.azure.functions version: 1.1.1

Please let me know if more details are required.

lilyjma:
@akshaykumars10 - SDK v1.3.0 is released! you can find it here [Central Repository: com/microsoft/durabletask-azure-functions/1.3.0 (apache.org)](https://repo.maven.apache.org/maven2/com/microsoft/durabletask-azure-functions/1.3.0/)

PR #153, opened 2023-07-12
Fix orchestration stuck using anyOf/allOf with retry policy

kaibocai:

resolves https://github.com/microsoft/durabletask-java/issues/149

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
@@ -970,6 +970,7 @@ final class TaskOrchestrationExecutor {
             private final TaskOrchestrationContext context;
             private final Instant firstAttempt;
             private final TaskFactory<V> taskFactory;
+            private Task<V> currentTask;
 
             private int attemptNumber;
             private FailureDetails lastFailure;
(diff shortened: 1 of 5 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -1026,6 +1042,8 @@ public V await() {
                     }
 
                     this.totalRetryTime  = Duration.between(startTime, this.context.getCurrentInstant());
+                    //empty currentTask.
+                    this.currentTask = null;
```
cgillum, line 1046: Rather than setting `this.currentTask = null` and then relying on the null-check at the top of the loop, wouldn't it be simpler to just reset the current task here and remove the null-check?

```java
// Generate a new task/future pair for the next attempt
this.currentTask = this.taskFactory.create();
setFuture(this.currentTask.future);
```

samples-azure-functions/src/main/java/com/functions/ParallelFunctions.java
```diff
@@ -0,0 +1,48 @@
...
+        String instanceId = client.scheduleNewOrchestrationInstance("Parallel");
+        context.getLogger().info("Created new Java orchestration with instance ID = " + instanceId);
+        return durableContext.createCheckStatusResponse(request, instanceId);
+    }
+
+    @FunctionName("Parallel")
+    public List<String> parallelOrchestrator(
+            @DurableOrchestrationTrigger(name = "ctx") TaskOrchestrationContext ctx) {
+        RetryPolicy retryPolicy = new RetryPolicy(2, Duration.ofSeconds(5));
+        TaskOptions taskOptions = new TaskOptions(retryPolicy);
```
cgillum, line 33: This test covers the base case where activities succeed, but shouldn't we also add a test for cases where at least one function call fails and is retried? Otherwise I feel like we're not really validating the correctness of this PR.

Also, would it make sense to have this test coverage in the main client library?

cgillum (commented):
Good fine. A few suggestions for improvements to this PR.

samples-azure-functions/src/main/java/com/functions/ParallelFunctions.java
kaibocai: > Also, would it make sense to have this test coverage in the main client library?

You mean add similar test cases for integration tests that run on durable sidecar right?

kaibocai:
> shouldn't we also add a test for cases where at least one function call fails and is retried

@cgillum, it turns out the current approach failed on this case. The thing is when customers using `anyOf/allOF` for example
```java
        tasks.add(ctx.callActivity("Append", "InputSad1", taskOptions, String.class));
        tasks.add(ctx.callActivity("Append", "InputSad1", taskOptions, String.class));
        tasks.add(ctx.callActivity("Append", "InputSad1", taskOptions, String.class));
        return ctx.allOf(tasks).await();
```
the `allOf` here returns a `CompletableTask` https://github.com/microsoft/durabletask-java/blob/b919929a0503f24ea5ac7a7e9d593a43a9dbcbdd/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java#L188, which is not retriable and we are awaiting on it. 
So all the `RetriableTask` tasks we put in the list won't be retried as the task we `await` on is a `CompletableTask`. 

So I am thinking let `allOf` return `CompletableTask` or `RetriableTask` depending on the tasks it has in the collection. However, there is another issue comes, as the collection of tasks can be mixed with `CompletableTask` and `RetriableTask`. In that case, I cannot run retry logic only on those `RetriableTask` in the collection. I can only run (or not run) retry logic on all of the tasks altogether. 
But maybe this is a corner case and it doesn't really matter to customers that few of their non-retriable tasks get retried? 

To really solve this issue, I think we should have a new GRPC type dedicate for `RetriableTask`, so the SDK can decide whether it should retry based on the message sent from the sidecar. 

cc: @davidmrdavid

cgillum:
> But maybe this is a corner case and it doesn't really matter to customers that few of their non-retriable tasks get retried?

I think this will be problematic in large fan-out cases. If 1/100 activities fails, the customer will expect that we only retry the 1 that failed and not the other 99. I don't see this as a corner case. I think we should take the time necessary to figure it out, though I'm okay with doing a partial fix since the current problem of getting stuck is worse than the problem of too much retrying.

> To really solve this issue, I think we should have a new GRPC type dedicate for RetriableTask, so the SDK can decide whether it should retry based on the message sent from the sidecar.

The gRPC layer has no notion of tasks, and this is by design. Tasks are purely an SDK concept, so there isn't any way to model it in the gRPC layer even if we wanted to. I can make time to try and help identify the correct design, but it will probably have to wait until after I get back from my vacation.

kaibocai:
Close this PR as a new one is created https://github.com/microsoft/durabletask-java/pull/157

PR #157, opened 2023-07-25
Re-implement retriable task

kaibocai:

resolves https://github.com/microsoft/durabletask-java/issues/149

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/Task.java
+++ b/client/src/main/java/com/microsoft/durabletask/Task.java
@@ -7,7 +7,6 @@ import com.microsoft.durabletask.interruption.OrchestratorBlockedException;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 import java.util.function.Function;
-import java.util.function.Supplier;
 
 /**
  * Represents an asynchronous operation in a durable orchestration.
@@ -32,9 +31,26 @@ import java.util.function.Supplier;
  */
 public abstract class Task<V> {
     final CompletableFuture<V> future;
-
+    private Task<V> ChildTask;
+    private Task<V> ParentTask;
     Task(CompletableFuture<V> future) {
         this.future = future;
+    }
+
+    public void setChildTask(Task<V> childTask) {
+        ChildTask = childTask;
+    }
+
+    public void setParentTask(Task<V> parentTask) {
+        ParentTask = parentTask;
+    }
+
+    public Task<V> getChildTask() {
+        return ChildTask;
+    }
+
+    public Task<V> getParentTask() {
+        return ParentTask;
     }
 
     /**
@@ -76,4 +92,10 @@ public abstract class Task<V> {
      * @return the new Task
      */
     public abstract Task<Void> thenAccept(Consumer<V> fn);
+
+    void notifyChildTaskCompletedSuccess(V result) {}
+
+    void notifyChildTaskCompletedExceptionally(Throwable ex) {}
+
+    void init() {}
 }
--- a/client/src/main/java/com/microsoft/durabletask/Task.java
+++ b/client/src/main/java/com/microsoft/durabletask/Task.java
@@ -31,26 +31,8 @@ import java.util.function.Function;
  */
 public abstract class Task<V> {
     final CompletableFuture<V> future;
-    private Task<V> childTask;
-    private Task<V> parentTask;
     Task(CompletableFuture<V> future) {
         this.future = future;
-    }
-
-    public void setChildTask(Task<V> childTask) {
-        this.childTask = childTask;
-    }
-
-    public void setParentTask(Task<V> parentTask) {
-        this.parentTask = parentTask;
-    }
-
-    public Task<V> getChildTask() {
-        return this.childTask;
-    }
-
-    public Task<V> getParentTask() {
-        return this.parentTask;
     }
 
     /**
--- a/client/src/main/java/com/microsoft/durabletask/Task.java
+++ b/client/src/main/java/com/microsoft/durabletask/Task.java
@@ -52,14 +52,6 @@ public abstract class Task<V> {
     }
 
     /**
-     * Returns {@code true} if the task was completed exceptionally.
-     * @return {@code true} if the task was completed exceptionally, otherwise {@code false}
-     */
-    public boolean isCompletedExceptionally() {
-        return this.future.isCompletedExceptionally();
-    }
-
-    /**
      * Blocks the orchestrator until this task to complete, and then returns its result.
      *
      * @return the result of the task
@@ -82,10 +74,4 @@ public abstract class Task<V> {
      * @return the new Task
      */
     public abstract Task<Void> thenAccept(Consumer<V> fn);
-
-    void handleChildSuccess(V result) {}
-
-    void handleChildException(Throwable ex) {}
-
-    void init() {}
 }
--- a/client/src/main/java/com/microsoft/durabletask/Task.java
+++ b/client/src/main/java/com/microsoft/durabletask/Task.java
@@ -70,6 +70,14 @@ public abstract class Task<V> {
     }
 
     /**
+     * Returns {@code true} if the task was completed exceptionally.
+     * @return {@code true} if the task was completed exceptionally, otherwise {@code false}
+     */
+    public boolean isCompletedExceptionally() {
+        return this.future.isCompletedExceptionally();
+    }
+
+    /**
      * Blocks the orchestrator until this task to complete, and then returns its result.
      *
      * @return the result of the task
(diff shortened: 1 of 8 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -988,45 +993,81 @@ private RetriableTask(
...
                 this.taskFactory = taskFactory;
                 this.policy = retryPolicy;
                 this.handler = retryHandler;
                 this.firstAttempt = context.getCurrentInstant();
                 this.totalRetryTime = Duration.ZERO;
+                this.constructParentChildTask(taskFactory);
+            }
+
+            // Every RetriableTask will hava a CompletableTask as child.
+            private void constructParentChildTask(TaskFactory<V> taskFactory) {
```
davidmrdavid, line 1006: can we call this `createChildTask`?

client/src/main/java/com/microsoft/durabletask/Task.java
```diff
@@ -76,4 +100,10 @@ public boolean isCancelled() {
      * @return the new Task
      */
     public abstract Task<Void> thenAccept(Consumer<V> fn);
+
+    void notifyChildTaskCompletedSuccess(V result) {}
+
+    void notifyChildTaskCompletedExceptionally(Throwable ex, Task<V> outerTask) {}
```
davidmrdavid, line 106: To simplify these names, can we call these
`handleChildSuccess`
and
`handleChildException`?

client/src/main/java/com/microsoft/durabletask/Task.java
```diff
@@ -32,11 +31,28 @@
...
+    }
+
+    public Task<V> getChildTask() {
+        return this.childTask;
+    }
+
+    public Task<V> getParentTask() {
+        return this.parentTask;
+    }
+
```
davidmrdavid, line 55: It seems the only class with "child tasks" are Retryable tasks. Therefore, I would recommend not to add `setChildTask` nor `getChildTask` to the `Task` class definition, as that's too general.

Can we add these methods to a more specific class or interface? I think should add them only to the `RetryableTask` class at this time.

davidmrdavid (commented):
Left some minor feedback.

Can we also add a test where we test that, when using `allOf`, a retryable task calling an always-failing activity will retry the activity the expected number of times?

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -1168,15 +1231,29 @@ public boolean isDone() {
...
-                return this.future.completeExceptionally(ex);
+                boolean result = false;
+                Task<V> parentTask = this.getParentTask();
+                if (parentTask != null) {
+                    // notify parent task
+                    parentTask.notifyChildTaskCompletedExceptionally(ex, this);
+                } else {
+                    result = this.future.completeExceptionally(ex);
+                }
+                return result;
```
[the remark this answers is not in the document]
kaibocai: This is the way I did at the beginning, but it caused the `await` on the normal `RetriableTask` failed, cause for the normal `RetriableTask` we are `await` on this future, however, we marked it as failed here `result = this.future.completeExceptionally(ex);` even though the later retry success. 

So to fix it, I pass the current task into `notifyChildTaskCompletedExceptionally` and mark it accordingly with the result of the `RetriableTask`

client/src/main/java/com/microsoft/durabletask/Task.java
kaibocai: So the child task of a `RetriableTask` is always a `CompletabeTask` and I need to maintain a `ParentTask` for this `CompletabeTask`, I just find it a little weird to move `ChildTask` to `RetriableTask` while left `ParentTask` in the `Task` interface.

kaibocai: Besides that, I am thinking in future maybe other types of `Task` can utilize `ChildTask` and `ParentTask`

davidmrdavid: I think it's definitely possible that, in the future, other types of `Task` may utilize `ChildTask` (they already utilize `ParentTask` so I think we should keep that here).

However, I think it would be better to wait until there's at least 2 classes that utilize `ChildTask` before placing those methods in this more general interface.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -988,45 +993,77 @@ private RetriableTask(
...
+                try{
+                    this.getChildTask().await();
+                } catch (OrchestratorBlockedException ex) {
+                    throw ex;
+                } catch (Exception ignored) {
+                    // ignore the exception from previous child tasks.
+                    // Only needs to return result from the last child task, which is on next line.
+                }
+                // Always return the last child task result.
+                return this.getChildTask().await();
```
davidmrdavid, line 1066: let's add some comments to explain this logic, since it's a bit fancy
```suggestion
                this.init();
                // when awaiting the first child task, we will continue iterating over the history until a result is found
                // for that task. If the result is an exception, the child task will invoke "handleChildException" on this
                // object, which awaits a timer, *re-sets the current child task to correspond to a retry of this task*,
                // and then awaits that child.
                // This logic continues until either the operation succeeds, or are our retry quota is met.
                // At that point, we break the `await()` on the child task.
                // Therefore, once we return from the following `await`,
                // we just need to await again on the *current* child task to obtain the result of this task
                try{
                    this.getChildTask().await();
                } catch (OrchestratorBlockedException ex) {
                    throw ex;
                } catch (Exception ignored) {
                    // ignore the exception from previous child tasks.
                    // Only needs to return result from the last child task, which is on next line.
                }
                // The child task at this point is correspond to the final result.  Either the task succeeded, or we can no longer retry it
                return this.getChildTask().await();
```

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -988,45 +993,77 @@ private RetriableTask(
...
+            void handleChildSuccess(V result) {
+                this.complete(result);
+            }
 
-                    this.attemptNumber++;
+            @Override
+            void handleChildException(Throwable ex) {
+                if (ex instanceof TaskFailedException) {
+                    tryRetry((TaskFailedException) ex);
+                }
```
davidmrdavid, line 1021: Should we throw an exception or log a warning if `ex` is not a `TaskFailedException`?

davidmrdavid (commented):
Almost there!

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
kaibocai: Good Catch, I review the code path again, I think for exception arrive here can only be `TaskFailedException`, so I plan to remove the if condition.

client/src/main/java/com/microsoft/durabletask/Task.java
```diff
@@ -53,6 +51,14 @@ public boolean isCancelled() {
...
     }
 
+    /**
+     * Returns {@code true} if the task was completed exceptionally.
+     * @return {@code true} if the task was completed exceptionally, otherwise {@code false}
+     */
+    public boolean isCompletedExceptionally() {
+        return this.future.isCompletedExceptionally();
+    }
+
```
davidmrdavid, line 61: I believe this method is unused

```suggestion
```

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -988,45 +995,91 @@ private RetriableTask(
...
                 this.context = context;
                 this.taskFactory = taskFactory;
                 this.policy = retryPolicy;
                 this.handler = retryHandler;
                 this.firstAttempt = context.getCurrentInstant();
                 this.totalRetryTime = Duration.ZERO;
+                this.createChildTask(taskFactory);
+            }
+
+            // Every RetriableTask will hava a CompletableTask as child.
```
davidmrdavid, line 1007:
```suggestion
            // Every RetriableTask will have a CompletableTask as a child task.
```

client/src/main/java/com/microsoft/durabletask/Task.java
```diff
@@ -76,4 +82,10 @@ public boolean isCancelled() {
      * @return the new Task
      */
     public abstract Task<Void> thenAccept(Consumer<V> fn);
+
+    void handleChildSuccess(V result) {}
+
+    void handleChildException(Throwable ex) {}
```
davidmrdavid, line 88: I thought we were moving these to `RetryableTask`?

kaibocai:
It turns out `anyOf` is not even working for the simplest scenario, I fix it to follow the same pattern we use for `allOf`.

davidmrdavid (approved):

PR #170, opened 2023-10-13
Fix log pipeline - populate logs into kusto

kaibocai:
With the current logging logic in the SDK, the logs are not recorded anywhere, causing the logs in the SDK to be meaningless. 

This PR updates the current Java Util Logger with `FunctionKustoHandler`, a customized `java.util.logging.Handler` that will stream the logs into kusto table. The picture below shows an example. 
image

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/log/FunctionKustoHandler.java
+++ b/client/src/main/java/com/microsoft/durabletask/log/FunctionKustoHandler.java
@@ -1,0 +1,37 @@
+package com.microsoft.durabletask.log;
+
+import java.io.PrintStream;
+import java.util.Arrays;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+class FunctionKustoHandler extends Handler {
+    private static final String FUNCTIONSKUSTOPREFIX = "LanguageWorkerConsoleLog";
+    private static final String DURABLEPREFIX = "DURABLE_JAVA_SDK";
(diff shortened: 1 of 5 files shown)


client/src/main/java/com/microsoft/durabletask/log/FunctionKustoHandler.java
```diff
@@ -0,0 +1,37 @@
+package com.microsoft.durabletask.log;
+
+import java.io.PrintStream;
+import java.util.Arrays;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+class FunctionKustoHandler extends Handler {
+    private static final String FUNCTIONSKUSTOPREFIX = "LanguageWorkerConsoleLog";
```
cgillum, line 10: Core logging abstractions can be in `durabletask-client` but anything specific to the either Kusto or Functions should instead go into the `durabletask-azure-functions` package. This is very important because `durabletask-client` will also be used outside of Azure by Dapr OSS customers, and we don't want to add any Azure-specific logic or dependencies for those users.

cgillum (changes requested):

client/src/main/java/com/microsoft/durabletask/log/FunctionKustoHandler.java
kaibocai: Got it.

Issue #169, opened 2023-10-11
Handlings errors from activity functions when fan-out/fan-in pattern is used


akshaykumars10:
We are implementing the fan-out/fan-in pattern in our use case. In this pattern,  If one of the activity function fails, we need to identify the specific function that encountered the error, in the orchestrator function.

Consider the following code block as an example:
`List<Integer> results = ctx.allOf(parallelTasks).await();`

In the event of a failure in any of the activity functions, the `results` variable remains unpopulated. This means that even if some activity functions have completed successfully, we are unable to distinguish them.

Is there a way to address this issue and accurately identify the successful activity functions?

kaibocai:
Hi @akshaykumars10 , for `allOf`, if there any exception happens for one of the activities, we will throw a `CompositeTaskFailedException` and it contains a list of exceptions that will contain all exceptions that happen during the process. 
You can inspect the exception list and check which task has failed. 
For example:
```java
    @FunctionName("Parallel")
    public List<String> parallelOrchestratorSad(
            @DurableOrchestrationTrigger(name = "ctx") TaskOrchestrationContext ctx,
            ExecutionContext context) {
        try {
            List<Task<String>> tasks = new ArrayList<>();
            tasks.add(ctx.callActivity("AppendSad", "Input1", String.class));
            tasks.add(ctx.callActivity("AppendHappy", "Input2", String.class));
            return ctx.allOf(tasks).await();
        } catch (CompositeTaskFailedException e) {
            for (Exception exception : e.getExceptions()) {
                if (exception instanceof TaskFailedException) {
                    TaskFailedException taskFailedException = (TaskFailedException) exception;
                    System.out.println("Task: " + taskFailedException.getTaskName() + " Failed for cause: " + taskFailedException.getErrorDetails().getErrorMessage());
                }
            }
        }
        return null;
    }

    @FunctionName("AppendHappy")
    public String appendHappy(
            @DurableActivityTrigger(name = "name") String name,
            final ExecutionContext context) {
        context.getLogger().info("AppendHappy: " + name);
        return name + "-test-happy";
    }

    @FunctionName("AppendSad")
    public String appendSad(
            @DurableActivityTrigger(name = "name") String name,
            final ExecutionContext context) {
        context.getLogger().info("Throw Test Exception: " + name);
        throw new RuntimeException("Test kaibocai exception");
    }
```
[...]

PR #174, opened 2023-10-19
Fix exception type when use RetriableTask in Fan in/out pattern

kaibocai:

Resolve https://github.com/microsoft/durabletask-java/issues/169

This PR fixes the exception type for using `RetriableTask` in `allOf` method. 

Description of the issue: 

For `RetriableTask` in `allOf` method if the task failed for exception, `CompositeTaskFailedException` is not thrown out instead it's a `TaskFailedException` being thrown out. The reason is that the `CompositeTaskFailedException` that built in [exceptionPath](https://github.com/microsoft/durabletask-java/blob/e938278aedbfadfd8b653bae9fe7afba30d05c18/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java#L202) haven't get chance to be obtained at [future.get()](https://github.com/microsoft/durabletask-java/blob/e938278aedbfadfd8b653bae9fe7afba30d05c18/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java#L1253), [ContextImplTask.this.processNextEvent()](https://github.com/microsoft/durabletask-java/blob/e938278aedbfadfd8b653bae9fe7afba30d05c18/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java#L1264C26-L1264C65) will first threw a `TaskFailedException`

For example:
```java
 @FunctionName("Parallel")
    public List<String> parallelOrchestratorSad(
            @DurableOrchestrationTrigger(name = "ctx") TaskOrchestrationContext ctx,
            ExecutionContext context) {
        try {
            List<Task<String>> tasks = new ArrayList<>();
            RetryPolicy policy = new RetryPolicy(2, Duration.ofSeconds(15));
            TaskOptions options = new TaskOptions(policy);
            tasks.add(ctx.callActivity("AppendSad", "Input1", options, String.class));
            tasks.add(ctx.callActivity("AppendHappy", "Input2", options, String.class));
            return ctx.allOf(tasks).await();
        } catch (CompositeTaskFailedException e) {
            // Nothing will be caught for this type of exception. 
            for (Exception exception : e.getExceptions()) {
                if (exception instanceof TaskFailedException) {
                    TaskFailedException taskFailedException = (TaskFailedException) exception;
                    System.out.println("Task: " + taskFailedException.getTaskName() + " Failed for cause: " + taskFailedException.getErrorDetails().getErrorMessage());
                }
            }
        }
        return null;
    }
    
    @FunctionName("AppendHappy")
    public String appendHappy(
            @DurableActivityTrigger(name = "name") String name,
            final ExecutionContext context) {
        context.getLogger().info("AppendHappy: " + name);
        return name + "-test-happy";
    }
    
    @FunctionName("AppendSad")
    public String appendSad(
            @DurableActivityTrigger(name = "name") String name,
            final ExecutionContext context) {
        context.getLogger().info("Throw Test Exception: " + name);
        throw new NullPointerException("Test kaibocai exception");
    }
```

This PR update the `do-while` block logics so that all exception will be handle inside the block https://github.com/microsoft/durabletask-java/blob/e938278aedbfadfd8b653bae9fe7afba30d05c18/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java#L1264

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
@@ -1279,10 +1279,8 @@ final class TaskOrchestrationExecutor {
                 } catch (Exception e) {
                     // ignore
                     /**
-                     * the idea is that for any exception thrown here it should be handled
-                     * in the do-while loop, the same exception should also be obtained when calling
-                     * {code#future.get()} in the do-while loop, and it will be handled correctly
-                     * in the catch block in the do-while loop.
+                     * We ignore the exception. Any Durable Task exceptions thrown here can be obtained when calling
+                     * {code#future.get()} in the implementation of 'await'. We defer to that loop to handle the exception.
                      */
                 }
                 // Any exception happen we return true so that we will enter to the do-while block for the last time.
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
@@ -1261,7 +1261,7 @@ final class TaskOrchestrationExecutor {
                             this.handleException(e);
                         }
                     }
-                } while (ContextImplTask.this.processNextEvent());
+                } while (processNextEvent());
 
                 // There's no more history left to replay and the current task is still not completed. This is normal.
                 // The OrchestratorBlockedException exception allows us to yield the current thread back to the executor so
@@ -1269,6 +1269,22 @@ final class TaskOrchestrationExecutor {
                 // This is *not* an exception - it's a normal part of orchestrator control flow.
                 throw new OrchestratorBlockedException(
                         "The orchestrator is blocked and waiting for new inputs. This Throwable should never be caught by user code.");
+            }
+
+            private boolean processNextEvent() {
+                try {
+                    return ContextImplTask.this.processNextEvent();
+                } catch (Exception e) {
+                    // ignore
+                    /**
+                     * the idea is that for any exception thrown here it should be handled
+                     * in the do-while loop, the same exception should also be obtained when calling
+                     * {code#future.get()} in the do-while loop, and it will be handled correctly
+                     * in the catch block in the do-while loop.
+                     */
+                }
+                // Any exception happen we return true so that we will enter to the do-while block for the last time.
+                return true;
             }
 
             @Override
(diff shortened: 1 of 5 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -1271,6 +1271,24 @@ public V await() {
                         "The orchestrator is blocked and waiting for new inputs. This Throwable should never be caught by user code.");
             }
 
+            private boolean processNextEvent() {
+                try {
+                    return ContextImplTask.this.processNextEvent();
+                } catch (OrchestratorBlockedException | ContinueAsNewInterruption exception) {
+                    throw exception;
```
davidmrdavid, line 1278: just to confirm: can processNextEvent actually lead to a ContinueAsNewInterruption? My understanding is that processNextEvent just iterates through the history, and I would expect that ContinueAsNewInterruption would only occur when the customer code is actually invoked.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -1271,6 +1271,24 @@ public V await() {
...
+                } catch (Exception e) {
+                    // ignore
+                    /**
+                     * the idea is that for any exception thrown here it should be handled
+                     * in the do-while loop, the same exception should also be obtained when calling
+                     * {code#future.get()} in the do-while loop, and it will be handled correctly
+                     * in the catch block in the do-while loop.
+                     */
+                }
+                // Any exception happen we return true so that we will enter to the do-while block for the last time.
```
davidmrdavid, line 1288: some nits on phrasing:

```suggestion
                    /**
                     * We ignore the exception. Any Durable Task exceptions thrown here can be obtained when calling
                     * {code#future.get()} in the implementation of 'await'. We defer to that loop to handle the exception.
                     */
                }
                // On an exception, we return true to the do-while block one last time.
```

davidmrdavid (approved):
LGTM so long as the suggested comment edit is incorporated.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
kaibocai: Yes, ContinueAsNewInterruption would only occur when the customer code is actually invoked. The SDK code will never cause a ContinueAsNewInterruption.

PR #176, opened 2023-10-24
Add deterministic GUID generation

kaibocai:
This is the work needed to allow orchestrations to generate random GUIDs in a deterministic way. It borrows from the similar logic in the durabletask-dotnet SDK.

Also this is request from https://github.com/dapr/java-sdk/issues/931

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
@@ -8,6 +8,7 @@ import java.time.Instant;
 import java.time.ZonedDateTime;
 import java.util.Arrays;
 import java.util.List;
+import java.util.UUID;
 
 /**
  * Used by orchestrators to perform actions such as scheduling tasks, durable timers, waiting for external events,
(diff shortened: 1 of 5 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
```diff
@@ -301,6 +302,20 @@ default void continueAsNew(Object input){
         this.continueAsNew(input, true);
     }
 
+    /**
+     * Create a new UUID that is safe for replay within an orchestration or operation.
```
cgillum, line 306: What do you mean by "or operation"?

cgillum (approved):
LGTM! Just one question about a doc comment, but otherwise I think this implementation looks good.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationContext.java
kaibocai: I am referencing to the doc add for durabletask-dotnet https://github.com/microsoft/durabletask-dotnet/blob/4889316ca9741dbaf62072f116b130921509f3e2/src/DurableTask/TaskOrchestrationContext.cs#L241
I think here it means any operation that uses the `newUUID` method and it is created inside the orchestration. :blush:

Issue #184, opened 2023-11-17
Deserialize Exception got swallowed when use `anyOf` with external event. 


kaibocai:
When use `anyOf` on waiting for an external event, a deserialize exception will be thrown out if the request body is not in JSON type or the Json style body cannot be correctly deserialized to POJO. However this exception seems to be swallowed with this fix https://github.com/microsoft/durabletask-java/pull/174, needs investigation.  open an issue for tracking this.

PR #185, opened 2023-11-20
Fix swallowed exception - Improve output exception message

kaibocai:

resolves https://github.com/microsoft/durabletask-java/issues/184. 

This PR fix the deserialized exception swallowed issue and improve the exception message in the final output to help customers better understand the exception. 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
+++ b/azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
@@ -9,7 +9,10 @@ package com.microsoft.durabletask.azurefunctions.internal.middleware;
 import com.microsoft.azure.functions.internal.spi.middleware.Middleware;
 import com.microsoft.azure.functions.internal.spi.middleware.MiddlewareChain;
 import com.microsoft.azure.functions.internal.spi.middleware.MiddlewareContext;
+import com.microsoft.durabletask.CompositeTaskFailedException;
+import com.microsoft.durabletask.DataConverter;
 import com.microsoft.durabletask.OrchestrationRunner;
+import com.microsoft.durabletask.TaskFailedException;
 import com.microsoft.durabletask.interruption.ContinueAsNewInterruption;
 import com.microsoft.durabletask.interruption.OrchestratorBlockedException;
 
@@ -48,6 +51,23 @@ public class OrchestrationMiddleware implements Middleware {
                 if (cause instanceof ContinueAsNewInterruption) {
                     throw (ContinueAsNewInterruption) cause;
                 }
+                // Below types of exception are raised by the client sdk, they data should be correctly pass back to
+                // durable function host. We need to cast them to the correct type so later when build the FailureDetails
+                // the correct exception data can be saved and pass back.
+                if (cause instanceof TaskFailedException) {
+                    throw (TaskFailedException) cause;
+                }
+
+                if (cause instanceof CompositeTaskFailedException) {
+                    throw (CompositeTaskFailedException) cause;
+                }
+
+                if (cause instanceof DataConverter.DataConverterException) {
+                    throw (DataConverter.DataConverterException) cause;
+                }
+                // e will be InvocationTargetException as using reflection, so we wrap it into a RuntimeException, so it
+                // won't change the current OrchestratorFunction API. We cannot throw the cause which is a Throwable, it
+                // requires update on OrchestratorFunction API.
                 throw new RuntimeException("Unexpected failure in the task execution", e);
             }
         });
--- a/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
+++ b/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
@@ -22,7 +22,6 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.IntFunction;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 final class TaskOrchestrationExecutor {
 
@@ -511,10 +510,13 @@ final class TaskOrchestrationExecutor {
                         rawResult != null ? rawResult : "(null)"));
 
             }
-
-            Object result = this.dataConverter.deserialize(rawResult, record.getDataType());
             CompletableTask task = record.getTask();
-            task.complete(result);
+            try {
+                Object result = this.dataConverter.deserialize(rawResult, record.getDataType());
+                task.complete(result);
+            } catch (Exception ex) {
+                task.completeExceptionally(ex);
+            }
         }
 
         private void handleTaskFailed(HistoryEvent e) {
@@ -558,11 +560,15 @@ final class TaskOrchestrationExecutor {
                 this.outstandingEvents.remove(eventName);
             }
             String rawResult = eventRaised.getInput().getValue();
-            Object result = this.dataConverter.deserialize(
-                    rawResult,
-                    matchingTaskRecord.getDataType());
             CompletableTask task = matchingTaskRecord.getTask();
-            task.complete(result);
+            try {
+                Object result = this.dataConverter.deserialize(
+                        rawResult,
+                        matchingTaskRecord.getDataType());
+                task.complete(result);
+            } catch (Exception ex) {
+                task.completeExceptionally(ex);
+            }
         }
 
         private void handleEventWhileSuspended (HistoryEvent historyEvent){
@@ -694,10 +700,13 @@ final class TaskOrchestrationExecutor {
                         rawResult != null ? rawResult : "(null)"));
 
             }
-
-            Object result = this.dataConverter.deserialize(rawResult, record.getDataType());
             CompletableTask task = record.getTask();
-            task.complete(result);
+            try {
+                Object result = this.dataConverter.deserialize(rawResult, record.getDataType());
+                task.complete(result);
+            } catch (Exception ex) {
+                task.completeExceptionally(ex);
+            }
         }
 
         private void handleSubOrchestrationFailed(HistoryEvent e){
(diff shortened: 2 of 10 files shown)


client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -558,11 +558,16 @@ private void handleEventRaised(HistoryEvent e) {
...
-                    matchingTaskRecord.getDataType());
+            Object result = null;
             CompletableTask task = matchingTaskRecord.getTask();
-            task.complete(result);
+            try {
+                result = this.dataConverter.deserialize(
+                        rawResult,
+                        matchingTaskRecord.getDataType());
+                task.complete(result);
+            } catch (DataConverter.DataConverterException ex) {
```
cgillum, line 568: Do we need the same exception handling logic in `handleTaskCompleted`, `handleSubOrchestrationCompleted`, and `waitForExternalEvent` as well?

kaibocai: Good catch! Will check each of them.

kaibocai: For `handleTaskCompleted` and `handleSubOrchestrationCompleted` we need same exception handling logic as they are called when we call `await` method somewhere. Whereas `waitForExternalEvent` it's called inside the orchestrator so it's exception handling logic is already correct at https://github.com/microsoft/durabletask-java/blob/8d3c1d4c46abc15725d8c56c836305c6a286cc0c/client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java#L60

After check all place calling `deserialize` method, it seems `handleEventRaised`, `handleTaskCompleted` and `handleSubOrchestrationCompleted` are the only three methods that need this exception handling logic

kaibocai: tag @cgillum

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
```diff
@@ -48,6 +51,17 @@ public void invoke(MiddlewareContext context, MiddlewareChain chain) throws Exce
                 if (cause instanceof ContinueAsNewInterruption) {
                     throw (ContinueAsNewInterruption) cause;
                 }
+                if (cause instanceof TaskFailedException) {
+                    throw (TaskFailedException) cause;
```
cgillum, line 55: The previous exceptions all have comments explaining that we're unwrapping `InvocationTargetException`. Is that also the case here? Should we update the comments in this catch block to explain this more wholistically? Otherwise, it's not easy to understand why you've added these new checks since there are no comments associated with them.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
cgillum: Cool. However, one more thing: in the latest iteration, you updated `handleTaskCompleted` and `handleSubOrchestrationCompleted` to catch `Exception`. However, here you are only catching `DataConverterException`. Should all of them be catching `Exception`?

samples-azure-functions/src/main/java/com/functions/DeserializeError.java
```diff
@@ -0,0 +1,109 @@
...
+import com.microsoft.durabletask.Task;
+import com.microsoft.durabletask.TaskOrchestrationContext;
+import com.microsoft.durabletask.azurefunctions.DurableActivityTrigger;
+import com.microsoft.durabletask.azurefunctions.DurableClientContext;
+import com.microsoft.durabletask.azurefunctions.DurableClientInput;
+import com.microsoft.durabletask.azurefunctions.DurableOrchestrationTrigger;
+
+import java.util.Optional;
+
+public class DeserializeError {
```
cgillum, line 20: The code under `samples-azure-functions/src/main/java/com/functions` is supposed to be references for customers to learn how to implement certain patterns. However, the functions in this file don't match that expectation because there is no error handling at all. I think we shouldn't be adding this file. I assume you added this to help with implementing the new tests in `EndToEndTests.java`. Can we instead target existing samples to try and trigger the error cases? If not, can those existing samples be updated to make it easier to write the negative end-to-end tests?

kaibocai: Distributed the error tests into corresponding samples for end-to-end testing.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
kaibocai: Added comments to explain this logic

samples-azure-functions/src/main/java/com/functions/AzureFunctions.java
```diff
@@ -55,4 +56,25 @@ public String capitalize(
...
+        DurableTaskClient client = durableContext.getClient();
+        String instanceId = client.scheduleNewOrchestrationInstance("DeserializeErrorOrchestrator");
+        context.getLogger().info("Created new Java orchestration with instance ID = " + instanceId);
+        return durableContext.createCheckStatusResponse(request, instanceId);
+    }
+
+    @FunctionName("DeserializeErrorOrchestrator")
+    public String deserializeErrorOrchestrator(
+            @DurableOrchestrationTrigger(name = "ctx") TaskOrchestrationContext ctx) {
+        // cause deserialize error
```
cgillum, line 76: I still don't feel good about these changes since they are effectively adding test code to the samples instead of adding real sample code. I feel like we should use a separate project for doing targeted end-to-end testing.

Alternatively, to hide the fact that we're using samples for error-handling tests, can we instead create a sample that demostrates how to work with POJOs? For example, an HTTP trigger could take a Person JSON object as an input, and the activity could be design to change it somehow (like capitalize the name) and then return the modified JSON object. To trigger a deserialization failure, you could simply send an invalid JSON payload to the HTTP trigger (which doesn't try to deserialize it) and then a failure can be triggered when the orchestrator tries to deserialize it. It's not perfect because it's not flexible enough to allow us to test different failure permutations, but I think it would be better than adding samples that are designed to fail (unless the sample can also include failure handling, in which case it might be a valuable sample to create).

kaibocai: I see the point. I agree we should separate the end2end tests and samples so I can test the failing scenarios (without error handling logic) instead of adding them to samples, which confuses customers. 
I will work on separating the tests and samples.

endtoendtests/src/test/java/com/functions/EndToEndTests.java
```diff
@@ -215,6 +215,53 @@ public void suspendResume() throws InterruptedException {
...
+            .statusCode(202);
+        // assert orchestration status
+        String statusQueryGetUri = jsonPath.get("statusQueryGetUri");
+        boolean completed = pollingCheck(statusQueryGetUri, "Failed", null, Duration.ofSeconds(10));
+        assertTrue(completed);
+
+        // assert exception message
+        Response resp = get(statusQueryGetUri);
+        String errorMessage = resp.jsonPath().get("output");
+        assertTrue(errorMessage.contains("Failed to deserialize the JSON text to java.lang.String"));
```
cgillum, line 242: One thing I'm curious about - is there no `failureDetails` property in the response? There probably should be. If not, we may want to consider opening an issue to expose this so that users can get more information about the failure, like the stack trace and the error type.

samples-azure-functions/src/main/java/com/functions/SubOrchestrator.java
```diff
@@ -0,0 +1,46 @@
...
+        DurableTaskClient client = durableContext.getClient();
+        String instanceId = client.scheduleNewOrchestrationInstance("CompletedErrorOrchestrator");
+        context.getLogger().info("Created new Java orchestration with instance ID = " + instanceId);
+        return durableContext.createCheckStatusResponse(request, instanceId);
+    }
+
+    @FunctionName("Orchestrator")
+    public String orchestrator(
+            @DurableOrchestrationTrigger(name = "ctx") TaskOrchestrationContext ctx) {
+        return ctx.callSubOrchestrator("CompletedErrorSubOrchestrator", "Austin", String.class).await();
```
cgillum, line 37: Is this correct?
```suggestion
        return ctx.callSubOrchestrator("SubOrchestrator", "Austin", String.class).await();
```

samples-azure-functions/src/main/java/com/functions/WaitExternalEvent.java
```diff
@@ -0,0 +1,42 @@
...
+        context.getLogger().info("Java HTTP trigger processed a request.");
+
+        DurableTaskClient client = durableContext.getClient();
+        String instanceId = client.scheduleNewOrchestrationInstance("ExternalEventActivity");
+        context.getLogger().info("Created new Java orchestration with instance ID = " + instanceId);
+        return durableContext.createCheckStatusResponse(request, instanceId);
+    }
+
+    @FunctionName("ExternalEventActivity")
+    public void externalEventActivity(@DurableOrchestrationTrigger(name = "runtimeState") TaskOrchestrationContext ctx)
```
cgillum, line 34: I see the method name is `externalEventActivity`, but should it instead be something like `externalEventOrchestrator`?

samples-azure-functions/src/main/java/com/functions/WaitExternalEvent.java
```diff
@@ -0,0 +1,42 @@
...
+    {
+        System.out.println("Waiting external event...");
+        Task<String> event = ctx.waitForExternalEvent("event", String.class);
+        Task<?> result = ctx.anyOf(event).await();
```
cgillum, line 38: As an example, it's a little strange to use `ctx.anyOf(...)` with just a single task. Should we improve this sample by adding a timer task (for example, 1 hour) or a second external event task ("event2")?

cgillum (commented):
Just a few small questions.

endtoendtests/src/test/java/com/functions/EndToEndTests.java
kaibocai: Yes from Java side at least there is only `output` and not `failureDetails`, I will open an issue to track it.

kaibocai: Issue created https://github.com/Azure/azure-functions-durable-extension/issues/2688
image

cgillum (approved):

PR #198, opened 2024-06-20
Add initial 1ES migration

davidmrdavid:
This PR adds an initial 1ES official pipeline. In order to do this, it removes the previous "manual" signing step in favor of the 1ES signing utils (which are yet to be added). Therefore, packages produced by this pipeline should not be released _yet_, at least not until we figure out how to incorporate the 1ES signing tool for Java.

### DIFF
--- a/azurefunctions/build.gradle
+++ b/azurefunctions/build.gradle
@@ -76,9 +76,10 @@ publishing {
     }
 }
 
-signing {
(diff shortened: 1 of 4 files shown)


kaibocai (approved):

kamperiadis (approved):
LGTM! I am assuming we will have the 1ES code signing step in a separate PR?

PR #249, opened 2025-12-15
Fixing gRPC channel shutdown

sophiatev:
This PR addresses an issue where a customer is running into an error when attempting to start a new orchestration:

`Previous channel ManagedChannelImpl{logId=57, target=localhost:4001} was not shutdown properly!!!`

The stack trace indicates that this occurs when attempting to build a new `managedSideCarChannel` in the `DurableTaskGrpcClient`. It appears that the previous `managedSidecarChannel` was not disposed properly in a call to `close()`. This can happen since we do nothing if the channel is not successfully terminated via a graceful call to `shutdown()` within the `timeout` (5 seconds). It can also happen if an `InterruptedException` is thrown, since we don't attempt any shutdown in that case.

To fix this, this PR initiates a forceful shutdown if the graceful shutdown fails in the timeframe. It also initiates a forceful shutdown in the case of an `InterruptedException` (but manually sets the thread's state to "interrupted" so that we do not lose this information). 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
@@ -71,11 +71,16 @@ public final class DurableTaskGrpcClient extends DurableTaskClient {
     public void close() {
         if (this.managedSidecarChannel != null) {
             try {
-                this.managedSidecarChannel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+                if (!this.managedSidecarChannel.shutdown().awaitTermination(5, TimeUnit.SECONDS))
+                {
+                    this.managedSidecarChannel.shutdownNow();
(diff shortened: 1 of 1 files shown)


sophiatev:
> Is there anyway to test or verify that this doesn't leave the system in a bad state? Even a test where we call `shutdownNow` and then recreate the connection and complete a request?

After spending much too long attempting to make this test, I think we are actually implicitly doing this for every test in the `IntegrationTests` class. This is because this [method](https://github.com/microsoft/durabletask-java/blob/31ead9c93633e974a4623f4bf18c6252cce5055e/client/src/test/java/com/microsoft/durabletask/IntegrationTestBase.java#L37) is called after each test which does a forceful shutdown on both gRPC channels being used.

halspang (approved):

PR #250, opened 2025-12-16
Release 1.6.2

sophiatev:
See title


azuremanaged/build.gradle
```diff
@@ -17,7 +17,7 @@ plugins {
 
 archivesBaseName = 'durabletask-azuremanaged'
 group 'com.microsoft'
-version = '1.6.1-preview.1'
+version = '1.6.2-preview.1'
```
halspang: Are we still technically in preview?

PR #263, opened 2026-02-23
Include Exception Properties at FailureDetails

nytian:
Adds support for inner failure details and custom exception properties in the Java SDK, matching the .NET SDK implementation ([durabletask-dotnet#474](https://github.com/microsoft/durabletask-dotnet/pull/474), [azure-functions-durable-extension#3215](https://github.com/Azure/azure-functions-durable-extension/pull/3215)).

- Add innerFailure and properties fields to FailureDetails, with full protobuf round-trip support
- Add ExceptionPropertiesProvider interface for custom property extraction via the worker builder
- Recursively capture inner exception chains from Throwable.getCause()
- Add unit tests and integration tests for the new functionality
- Backward compatible — existing users who don't register a provider see no behavior change.


client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -20,29 +25,76 @@ public final class FailureDetails {
...
+
+    /**
+     * Creates a {@code FailureDetails} from an exception, optionally using the provided
+     * {@link ExceptionPropertiesProvider} to extract custom properties.
+     *
+     * @param exception the exception that caused the failure
+     * @param provider  the provider for extracting custom properties, or {@code null}
+     * @return a new {@code FailureDetails} instance
+     */
+    static FailureDetails fromException(Exception exception, @Nullable ExceptionPropertiesProvider provider) {
```
bachuv: Is there a way to avoid duplicate code and have `fromException()` call `fromExceptionRecursive()`?

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -124,10 +203,98 @@ static String getFullStackTrace(Throwable e) {
...
+
+        if (this.properties != null) {
+            builder.putAllProperties(convertToProtoProperties(this.properties));
+        }
+
+        return builder.build();
+    }
+
+    @Nullable
+    private static FailureDetails fromExceptionRecursive(
```
bachuv, line 228: Should we add a max depth limit along with the `exception == null` check to avoid cyclic and deep exception chains?

client/build.gradle
bachuv, line 1: Can we also add examples of how to use this in the samples directory?

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -86,6 +138,28 @@ public boolean isNonRetriable() {
...
+
+    /**
+     * Gets additional properties associated with the exception, or {@code null} if no properties are available.
+     * <p>
+     * The returned map is unmodifiable.
+     *
+     * @return an unmodifiable map of property names to values, or {@code null}
+     */
+    @Nullable
+    public Map<String, Object> getProperties() {
```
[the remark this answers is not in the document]
nytian: the created properties is already immutable

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -86,6 +124,31 @@ public boolean isNonRetriable() {
...
+        if (this.properties == null) {
+            return null;
+        }
+        return Collections.unmodifiableMap(this.properties);
```
bachuv: Can we just return `this.properties` here? We already have this line in the constructor: `this.properties = properties != null ? Collections.unmodifiableMap(new HashMap<>(properties)) : null;`

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -20,29 +25,62 @@ public final class FailureDetails {
...
         this.stackTrace = errorDetails;
 
         // Error message can be null for things like NullPointerException but the gRPC contract doesn't allow null
         this.errorMessage = errorMessage != null ? errorMessage : "";
         this.isNonRetriable = isNonRetriable;
+        this.innerFailure = innerFailure;
+        this.properties = properties != null ? Collections.unmodifiableMap(new HashMap<>(properties)) : null;
     }
 
     FailureDetails(Exception exception) {
```
bachuv: Is `FailureDetails(Exception exception)` used anywhere?

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -20,29 +25,62 @@ public final class FailureDetails {
...
+        return fromExceptionRecursive(exception, provider, 0);
     }
 
     FailureDetails(TaskFailureDetails proto) {
         this(proto.getErrorType(),
              proto.getErrorMessage(),
              proto.getStackTrace().getValue(),
-             proto.getIsNonRetriable());
+             proto.getIsNonRetriable(),
+             proto.hasInnerFailure() ? new FailureDetails(proto.getInnerFailure()) : null,
```
bachuv, line 83: Should we add a depth limit for this recursive call too?

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
bachuv, line 179: isCausedBy() only checks the top-level errorType, but now that innerFailure exists, users might expect it to traverse the chain. Can we either update the Javadoc to clarify this or traverse the chain? Can we also check what the behavior is in .NET and match it?

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -124,10 +180,126 @@ static String getFullStackTrace(Throwable e) {
...
+                for (Value item : value.getListValue().getValuesList()) {
+                    list.add(convertProtoValue(item));
+                }
+                return list;
+            case STRUCT_VALUE:
+                Map<String, Object> map = new HashMap<>();
+                for (Map.Entry<String, Value> entry : value.getStructValue().getFieldsMap().entrySet()) {
+                    map.put(entry.getKey(), convertProtoValue(entry.getValue()));
+                }
+                return map;
```
bachuv: getProperties() returns an unmodifiable map, but if a property value is a List or nested Map (from proto LIST_VALUE/STRUCT_VALUE), those inner collections are plain ArrayList/HashMap then callers can mutate them. Consider wrapping them with Collections.unmodifiableList()/Collections.unmodifiableMap() in convertProtoValue for consistency.

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -124,10 +180,126 @@ static String getFullStackTrace(Throwable e) {
...
+            int depth) {
+        if (exception == null || depth > MAX_INNER_FAILURE_DEPTH) {
+            return null;
+        }
+        Map<String, Object> properties = null;
+        if (provider != null && exception instanceof Exception) {
+            try {
+                properties = provider.getExceptionProperties((Exception) exception);
+            } catch (Exception ignored) {
+                // Don't let provider errors mask the original failure
```
bachuv: Should we add a log here to provide more information about the exception?

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -45,7 +46,7 @@ public TaskOrchestrationExecutor(
             Duration maximumTimerInterval,
             Logger logger,
             DurableTaskGrpcWorkerVersioningOptions versioningOptions) {
-        this(orchestrationFactories, dataConverter, maximumTimerInterval, logger, versioningOptions, false);
+        this(orchestrationFactories, dataConverter, maximumTimerInterval, logger, versioningOptions, false,null);
```
bachuv: nit
```suggestion
        this(orchestrationFactories, dataConverter, maximumTimerInterval, logger, versioningOptions, false, null);
```

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
nytian: update the doc. .NET SDK TaskFailureDetails.IsCausedBy there also only inspects  ErrorType and does not walk InnerFailure ([code](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/TaskFailureDetails.cs)). To stay consistent across language sdk just updated the doc

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
nytian: I think we are fine with it.  The producer side (fromExceptionRecursive here, and the .NET equivalent) already enforces MAX_INNER_FAILURE_DEPTH = 10, and gRPC's max message size caps it further ( as recursive call will make the message size grows larger and larger), so we shouldn't see a chain deep enough to be a stack risk on this side.

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -389,11 +392,9 @@ public void startAndBlock() {
                                 activityRequest.getTaskId());
                         } catch (Throwable e) {
                             activityError = e;
-                            failureDetails = TaskFailureDetails.newBuilder()
-                                .setErrorType(e.getClass().getName())
-                                .setErrorMessage(e.getMessage())
-                                .setStackTrace(StringValue.of(FailureDetails.getFullStackTrace(e)))
-                                .build();
+                            Exception ex = e instanceof Exception ? (Exception) e : new RuntimeException(e);
```
bachuv: Why do we default to RuntimeException()? In Java, throwables have two types:

```
Throwable
├── Exception   (recoverable — e.g. IOException, IllegalArgumentException)
└── Error       (serious — e.g. OutOfMemoryError, StackOverflowError)
```

The activity catch block catches `Throwable` (both branches), but `fromException()` only accepts `Exception`. So when an `Error` is caught, the code wraps it:

```java
Exception ex = e instanceof Exception ? (Exception) e : new RuntimeException(e);
```

This changes what gets reported:

| | errorType | errorMessage |
|---|---|---|
| **Before** | `java.lang.StackOverflowError` | `null` |
| **After** | `java.lang.RuntimeException` | `"java.lang.StackOverflowError"` |

The real error type is hidden behind the wrapper. This only affects `Error` subclasses — regular exceptions are unaffected.

Can we have `fromException` accept `Throwable` directly (it already recurses into `getCause()` which returns `Throwable`) and only invoke the `ExceptionPropertiesProvider` when the throwable is an `Exception`, since the provider API is typed to `Exception`? What are your thoughts on this?

bachuv (approved):

Issue #27, opened 2022-04-06
Build and publish pipeline for Maven packages


cgillum:
Work needs to be done to publish packages to the maven central repository: https://mvnrepository.com/. This includes both the annotations package (currently named `com.microsoft.durabletask.azurefunctions`) and the core SDK package (`com.microsoft.durabletask`)

shreyas-gopalakrishna:
Documenting the steps to follow to release the durabletask client and durabletask annotations packages.

### Build the project and install to local maven repository
Update the package versions in the build.gradle files for both the packages
    - [durabletask-java client](https://github.com/microsoft/durabletask-java/blob/main/client/build.gradle#L10)
    - [durabletask-java annotations](https://github.com/microsoft/durabletask-java/blob/main/azurefunctions/build.gradle)

Configure local system to use JDK 8. Verify using the below command before proceeding
    - `java -version`

Build and install the durable task java client to local maven repository using IntelliJ or the below commands 
    - `gradle client:build`
    - `gradle client:publishToMavenLocal`

Build and install the durable task java annotations to local maven repository using IntelliJ or the below commands 
    - `gradle azurefunctions:build`
    - `gradle azurefunctions:publishToMavenLocal`

The updated jars can be located at the local maven repository in the below path.
    - `C:\Users\<user>\.m2\repository\com\microsoft\durabletask-client\<version>`
    - `C:\Users\<user>\.m2\repository\com\microsoft\durabletask-azure-functions<version>`


### Publish packages to maven
The [Partner Release Pipeline](https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/1/Partner-Release-Pipeline) will be used to publish packages into sonatype and maven.

Upload the files from your local maven repository to the [Azure drops container](https://ms.portal.azure.com/#blade/Microsoft_Azure_Storage/ContainerMenuBlade/overview/storageAccountId/%2Fsubscriptions%2Fa18897a6-7e44-457d-9260-f2854c0aca42%2FresourceGroups%2FAzureSDKEng%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Fazuresdkpartnerdrops/path/drops/etag/%220x8D680AE5A72EEFF%22) 
    - Create a new folder for your package (if a new package) and a version sub folder.
    - Container locations `/azure-functions/java/durabletask-azure-functions/<version>` and `/azure-functions/java/durabletask-client/<version>`
[...]

Issue #11, opened 2022-01-31
Query instance history support


cgillum:
### Requirement

Orchestration clients should be able to query for an orchestration instance and also get back the history events for that orchestration, similar to the `showHistory` flag mentioned [here](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management?tabs=csharp#query-instances).

### Implementation notes

The .NET in-process Durable Functions implementation returned a JSON array. For this work-item, we should instead return something structured - e.g., an array of strongly typed objects.

PR #268, opened 2026-03-04
Durable Entities

### LINKED ISSUE #31: Entity support
For feature parity, durable entities should be supported at some point.

bachuv:

Adding Durable Entities support for the Java SDK.

resolves #31 

### Pull request checklist


client/src/main/java/com/microsoft/durabletask/DurableEntityClient.java
```diff
@@ -0,0 +1,232 @@
...
+     * }
+     * }</pre>
+     *
+     * @param query     the query filter criteria
+     * @param stateType the expected type of the entity's state, used with
+     *                  {@link EntityMetadata#readStateAs(Class)} for deserialization
+     * @param <T>       the entity state type
+     * @return a pageable iterable over all matching entities with state included
+     */
+    public <T> EntityQueryPageable getAllEntities(EntityQuery query, Class<T> stateType) {
```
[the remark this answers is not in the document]
bachuv: Fixed

samples-azure-functions/src/main/java/com/functions/entities/CounterFunctions.java
```diff
@@ -0,0 +1,207 @@
...
+     * Signals the counter entity to reset to zero.
+     */
+    @FunctionName("Counter_Reset")
+    public HttpResponseMessage counterReset(
+            @HttpTrigger(name = "req", methods = {HttpMethod.POST},
+                    route = "counters/{id}/reset",
+                    authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            @BindingName("id") String id,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
bachuv: Fixed.

samples-azure-functions/src/main/java/com/functions/entities/LifetimeFunctions.java
```diff
@@ -0,0 +1,108 @@
...
+     * operation; otherwise uses the standard {@code delete} operation.
+     */
+    @FunctionName("Lifetime_Delete")
+    public HttpResponseMessage lifetimeDelete(
+            @HttpTrigger(name = "req", methods = {HttpMethod.DELETE},
+                    route = "lifetimes/{id}",
+                    authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            @BindingName("id") String id,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
bachuv: Fixed

samples-azure-functions/src/main/java/com/functions/entities/CounterFunctions.java
```diff
@@ -0,0 +1,207 @@
...
+     * Deletes the counter entity using the built-in implicit "delete" operation.
+     */
+    @FunctionName("Counter_Delete")
+    public HttpResponseMessage counterDelete(
+            @HttpTrigger(name = "req", methods = {HttpMethod.DELETE},
+                    route = "counters/{id}",
+                    authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            @BindingName("id") String id,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
bachuv: Fixed

samples-azure-functions/src/main/java/com/functions/entities/CounterFunctions.java
```diff
@@ -0,0 +1,207 @@
...
+     * Gets the current state of the counter entity.
+     */
+    @FunctionName("Counter_Get")
+    public HttpResponseMessage counterGet(
+            @HttpTrigger(name = "req", methods = {HttpMethod.GET},
+                    route = "counters/{id}",
+                    authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            @BindingName("id") String id,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
bachuv: Fixed.

samples-azure-functions/src/main/java/com/functions/entities/LifetimeFunctions.java
```diff
@@ -0,0 +1,108 @@
...
+     * Initializes the lifetime entity by sending an "init" signal.
+     */
+    @FunctionName("Lifetime_Init")
+    public HttpResponseMessage lifetimeInit(
+            @HttpTrigger(name = "req", methods = {HttpMethod.PUT},
+                    route = "lifetimes/{id}",
+                    authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            @BindingName("id") String id,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
bachuv: Fixed

samples-azure-functions/src/main/java/com/functions/entities/LifetimeFunctions.java
```diff
@@ -0,0 +1,108 @@
...
+     * Gets the current state of the lifetime entity.
+     */
+    @FunctionName("Lifetime_Get")
+    public HttpResponseMessage lifetimeGet(
+            @HttpTrigger(name = "req", methods = {HttpMethod.GET},
+                    route = "lifetimes/{id}",
+                    authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            @BindingName("id") String id,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
bachuv: Fixed

samples-azure-functions/src/main/java/com/functions/entities/UserFunctions.java
```diff
@@ -0,0 +1,228 @@
...
+     * Deletes the user entity using the implicit "delete" operation from {@code TaskEntity}.
+     */
+    @FunctionName("DeleteUser")
+    public HttpResponseMessage deleteUser(
+            @HttpTrigger(name = "req", methods = {HttpMethod.DELETE},
+                    route = "users/{id}",
+                    authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            @BindingName("id") String id,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
bachuv: Fixed

samples-azure-functions/src/main/java/com/functions/entities/UserFunctions.java
```diff
@@ -0,0 +1,228 @@
...
+     * Gets the current state of the user entity.
+     */
+    @FunctionName("GetUser")
+    public HttpResponseMessage getUser(
+            @HttpTrigger(name = "req", methods = {HttpMethod.GET},
+                    route = "users/{id}",
+                    authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
+            @DurableClientInput(name = "durableContext") DurableClientContext durableContext,
+            @BindingName("id") String id,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
bachuv: Fixed

client/src/main/java/com/microsoft/durabletask/TaskEntity.java
```diff
@@ -0,0 +1,430 @@
...
+                if (m.getDeclaringClass() == TaskEntity.class) {
+                    continue;
+                }
+                // Skip methods from ITaskEntity interface
+                if (m.getDeclaringClass() == ITaskEntity.class) {
+                    continue;
+                }
+                // Skip methods from JDK packages to prevent unintended state dispatch
+                // (e.g., Integer.intValue(), String.length()) when state type is a JDK class
+                String declaringPackage = m.getDeclaringClass().getPackageName();
```
YunchuWang: **Bug (Critical): `Class.getPackageName()` is Java 9+ but the project targets Java 8.**

The build config (`client/build.gradle`) sets `sourceCompatibility = JavaVersion.VERSION_1_8`. `Class.getPackageName()` was introduced in Java 9 and will throw `NoSuchMethodError` at runtime on Java 8.

Suggested fix: use `Class.getPackage().getName()` instead, which is available since Java 1.0:
```java
Package pkg = m.getDeclaringClass().getPackage();
String declaringPackage = pkg != null ? pkg.getName() : "";
```
Note: `getPackage()` can return `null` for dynamically generated classes, so a null check is needed.

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -70,6 +77,11 @@ public final class DurableTaskGrpcWorker implements AutoCloseable {
         this.dataConverter = builder.dataConverter != null ? builder.dataConverter : new JacksonDataConverter();
         this.maximumTimerInterval = builder.maximumTimerInterval != null ? builder.maximumTimerInterval : DEFAULT_MAXIMUM_TIMER_INTERVAL;
         this.versioningOptions = builder.versioningOptions;
+        this.workItemExecutor = Executors.newCachedThreadPool(r -> {
```
YunchuWang: **Bug (Critical): `maxConcurrentEntityWorkItems` is never enforced.**

The builder accepts `maxConcurrentEntityWorkItems` (default 1) and stores it on the worker, but the thread pool is `Executors.newCachedThreadPool()` which is unbounded. There's no semaphore, bounded pool, or any other mechanism that actually limits concurrent entity work items.

Suggested fix: Use a `Semaphore` to gate entity work item submission, or use `Executors.newFixedThreadPool(maxConcurrentEntityWorkItems)` for the entity work items (keeping a separate pool or shared pool for activities if needed).

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -340,42 +371,142 @@ public void startAndBlock() {
...
-                            failureDetails = TaskFailureDetails.newBuilder()
-                                .setErrorType(e.getClass().getName())
-                                .setErrorMessage(e.getMessage())
-                                .setStackTrace(StringValue.of(FailureDetails.getFullStackTrace(e)))
-                                .build();
-                        } finally {
-                            activityScope.close();
-                            TracingHelper.endSpan(activitySpan, activityError);
-                        }
+                        this.workItemExecutor.submit(() -> {
```
YunchuWang: **Bug (Medium): Activity execution silently changed from synchronous to concurrent.**

Previously, activity tasks were executed synchronously in the `getWorkItems` loop (one at a time). This PR wraps them in `workItemExecutor.submit()`, making them run concurrently with no upper bound (since `newCachedThreadPool` is used).

This is a significant behavioral change that affects all users, not just entity users. Was this intentional? If so, it should be:
1. Called out in the PR description and CHANGELOG
2. Controllable via a builder option (e.g., `maxConcurrentActivityWorkItems`)
3. Separate from the entity concurrency control

If unintentional, activity execution should remain synchronous or be gated properly.

client/src/main/java/com/microsoft/durabletask/GrpcDurableEntityClient.java
```diff
@@ -0,0 +1,175 @@
...
+                continuationToken,
+                totalEmptyEntitiesRemoved,
+                totalOrphanedLocksReleased);
+    }
+
+    private EntityMetadata toEntityMetadata(
+            com.microsoft.durabletask.implementation.protobuf.OrchestratorService.EntityMetadata protoEntity) {
+        Instant lastModifiedTime = DataConverter.getInstantFromTimestamp(protoEntity.getLastModifiedTime());
+        String lockedBy = protoEntity.hasLockedBy() ? protoEntity.getLockedBy().getValue() : null;
+        String serializedState = protoEntity.hasSerializedState()
```
YunchuWang, line 169: **Bug (Medium): `includesState` is derived from `hasSerializedState()` which conflates "state was requested" with "state physically exists".**

If a caller queries with `includeState=true` but the entity has null/empty state, `hasSerializedState()` returns `false` and `isIncludesState()` will return `false`. This misleads callers who use `isIncludesState()` to determine whether they need to re-query with state.

The `includesState` flag should reflect the *request's* `includeState` setting, not the proto field presence. Consider passing the query's `includeState` flag through to this method:
```java
return new EntityMetadata(
    protoEntity.getInstanceId(),
    lastModifiedTime,
    protoEntity.getBacklogQueueSize(),
    lockedBy,
    serializedState,
    includeState,  // from the query/request, not proto field presence
    this.dataConverter);
```

client/src/main/java/com/microsoft/durabletask/EntityQueryPageable.java
```diff
@@ -0,0 +1,180 @@
...
+
+            if (continuationToken == null || continuationToken.isEmpty()) {
+                finished = true;
+            }
+
+            return result;
+        }
+    }
+
+    private static EntityQuery cloneQuery(EntityQuery source) {
```
YunchuWang: **Code Quality: `cloneQuery` duplicates `EntityQuery.copy()` and double-normalizes.**

`EntityQuery` already has a `copy()` method that does exactly this. Using `cloneQuery` instead introduces:
1. A maintenance burden (two places to update when fields are added)
2. Double-normalization: passing an already-normalized `@name@key` value through `setInstanceIdStartsWith()` re-runs the normalization logic (re-lowercasing, `@` detection). While currently harmless (idempotent), it's fragile.

Suggested fix: replace `cloneQuery(baseQuery)` with `baseQuery.copy()` and remove this private method.

client/src/main/java/com/microsoft/durabletask/TaskEntityOperation.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Creates a new {@code TaskEntityOperation}.
+     *
+     * @param name            the operation name
+     * @param serializedInput the serialized input data, or {@code null} if no input
+     * @param context         the entity context
+     * @param state           the entity state
+     * @param dataConverter   the data converter for deserializing input
+     */
+    public TaskEntityOperation(
```
YunchuWang: **API Design: Constructor should be package-private, not `public`.**

`TaskEntityOperation` is created exclusively by the framework (`TaskEntityExecutor`). Making the constructor `public` allows external code to create instances with arbitrary state, which could lead to misuse. Consider removing the `public` modifier to restrict it to package-private access, consistent with other framework-internal constructors like `CleanEntityStorageResult`.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -63,7 +81,11 @@ public TaskOrchestratorResult execute(
             while (context.processNextEvent()) { /* no method body */ }
             completed = true;
         } catch (OrchestratorBlockedException orchestratorBlockedException) {
-            logger.fine("The orchestrator has yielded and will await for new events.");
+            logger.info(String.format(
```
YunchuWang: **Operational: Log level changed from `fine` to `info` for orchestrator yield message.**

This message fires every time an orchestrator yields (which is the normal happy path for most orchestrations). At `info` level, this will produce significant log volume in production. The previous `fine` level was appropriate.

Similarly, the new event processing log at line ~1604 (`Processing new event`) logs at `info` for *every* history event during replay. For orchestrations with many events, this multiplies log volume dramatically.

Suggested fix: revert to `fine`/`FINE` level for all replay-path logging. Reserve `info` for genuinely noteworthy events (entity operation failures, unexpected conditions).

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -340,42 +371,142 @@ public void startAndBlock() {
...
+                                        String.format("Failed to execute entity batch for '%s'. Abandoning work item.",
+                                                entityRequest.getInstanceId()),
+                                        e);
+                                this.sidecarClient.abandonTaskEntityWorkItem(
+                                        AbandonEntityTaskRequest.newBuilder()
+                                                .setCompletionToken(workItem.getCompletionToken())
+                                                .build());
+                            }
+                        });
+                    } else if (requestType == RequestCase.ENTITYREQUESTV2) {
```
YunchuWang, line 436: **Code Quality: Inconsistent indentation in ENTITYREQUESTV2 handler.**

The code inside this lambda has inconsistent indentation compared to the V1 (ENTITYREQUEST) handler above. For example, the `EntityBatchRequest.Builder` line is indented with spaces that don't align with the enclosing `try` block. Please re-indent the lambda body to match the style of the V1 handler.

client/src/main/java/com/microsoft/durabletask/CleanEntityStorageRequest.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Gets whether the client should automatically continue cleaning until all entities are processed.
+     *
+     * @return {@code true} if the client will automatically continue until complete
+     */
+    public boolean isContinueUntilComplete() {
+        return this.continueUntilComplete;
+    }
+
+    private boolean continueUntilComplete = false;
```
YunchuWang: **Code Quality: Field `continueUntilComplete` is declared after the methods that reference it.**

All other fields in this class (and in other classes in this PR) are declared at the top of the class before any methods. This field should be moved up to join the other field declarations near line 17 for consistency.

client/src/main/java/com/microsoft/durabletask/TaskEntity.java
```diff
@@ -0,0 +1,430 @@
...
+
+    /**
+     * Finds a public method on the target class matching the operation name (case-insensitive).
+     * Methods inherited from {@code Object} and from {@code TaskEntity} itself are excluded.
+     */
+    @Nullable
+    private static Method findMethod(Class<?> targetClass, String operationName) {
+        Map<String, Optional<Method>> classCache = methodCache.get(targetClass);
+        String key = operationName.toLowerCase();
+        return classCache.computeIfAbsent(key, k -> findMethodUncached(targetClass, k)).orElse(null);
```
YunchuWang: **Code Quality: Inconsistent indentation in `findMethodUncached`.**

The method body has extra leading whitespace (12 spaces instead of 8). The `List<Method> matches = ...` line and subsequent code should be indented at the same level as other private methods in this class.

YunchuWang (changes requested):
## Code Review: Durable Entities for Java SDK

Thorough review of the 78-file PR adding Durable Entities support. The overall design is solid and well-documented, with good alignment to the .NET SDK patterns. However, I found several issues ranging from a **Java 8 compatibility breakage** to unused concurrency controls and excessive logging. Details in inline comments below.

### Summary of Findings

**Critical (must fix):**
1. `Class.getPackageName()` breaks Java 8 compatibility (project targets Java 8)
2. `maxConcurrentEntityWorkItems` is accepted but never enforced
3. Activity execution silently changed from synchronous to unbounded-concurrent

**Medium:**
4. `EntityMetadata.includesState` incorrectly derived from proto field presence
5. `EntityQueryPageable.cloneQuery()` duplicates `EntityQuery.copy()` and double-normalizes
6. `TaskEntityOperation` constructor should be package-private

**Low / Code Quality:**
7. INFO-level logging on orchestration replay hot path will flood production logs
8. Inconsistent indentation in ENTITYREQUESTV2 handler and `findMethodUncached`
9. Field declared after methods that reference it in `CleanEntityStorageRequest`
10. Missing newline at end of several files

YunchuWang (changes requested):
## 🔍 Strict Code Review — Durable Entities (Java SDK)

This is a thorough review of the 12,394-line, 78-file Durable Entities feature addition. The overall architecture is solid and well-aligned with the .NET SDK design. However, there are several correctness, safety, and API-design issues that should be addressed before merge.

I've organized findings into tiers — please prioritize the **Critical/High** items.

---

### 🔴 Critical / High Priority

**1. Lock release `close()` is not idempotent — can cause non-deterministic replay (`TaskOrchestrationExecutor.java`)**
The `AutoCloseable` returned from `lockEntities()` sends unlock messages and increments `sequenceNumber` every time `close()` is called. If a user calls `close()` twice (very common with try-with-resources + explicit close, or exception handling double-close), this will: (a) send duplicate unlock messages corrupting entity state, and (b) increment `sequenceNumber` again, causing **non-deterministic replay failures** on subsequent executions. **Fix:** Add a boolean guard to make `close()` idempotent.

**2. Critical section enforcement bypassed when `lockedEntityIds` is null (`TaskOrchestrationExecutor.java`)**
In `callEntity()`:
```java
if (this.isInCriticalSection && this.lockedEntityIds != null && !this.lockedEntityIds.contains(...))
```
`handleEntityLockGranted()` sets `this.lockedEntityIds = this.pendingLockSets.remove(criticalSectionId)`, which returns `null` if the key is missing. Combined with `isInCriticalSection = true`, the null short-circuits the safety check, allowing calls to **any** entity — defeating deadlock prevention. **Fix:** Throw if `pendingLockSets.remove()` returns null.

**3. Activity completion failures silently lost (`DurableTaskGrpcWorker.java`)**
Activities are now dispatched to `workItemExecutor`, but `sidecarClient.completeActivityTask()` inside the `Runnable` has no outer try-catch for the gRPC call itself. If the gRPC call fails (network issue, sidecar crash), the exception hits only the thread pool's uncaught exception handler — the activity result is **permanently lost** with no retry or abandon. Entity work items have an abandon fallback, but activities do not.

**4. `setState()` silently deletes entity state on serialization failure (`TaskEntityState.java`)**
```java
this.serializedState = this.dataConverter.serialize(state);
if (this.serializedState == null) {
    deleteState(); // <-- silent data loss!
}
```
Since the outer `if` already guarantees `state != null`, `serialize()` returning null means a broken `DataConverter`. Silently calling `deleteState()` is a **data-loss bug**. Should throw `IllegalStateException`.

**5. Unbounded thread pool (`DurableTaskGrpcWorker.java`)**
`Executors.newCachedThreadPool()` has no upper bound. Despite `maxConcurrentEntityWorkItems` being sent to the sidecar, the local thread pool is unbounded. Activities AND entities both use this pool. A burst of work items can create unbounded threads. Use a bounded `ThreadPoolExecutor` or separate pools.

**6. Activity execution model silently changed from sync to async (`DurableTaskGrpcWorker.java`)**
Activities were previously executed synchronously on the main dispatch thread. Moving them to the thread pool is a **silent behavioral breaking change**: (a) activities using `ThreadLocal` state will break, (b) activities relying on sequential execution relative to dispatch will break, (c) gRPC context propagation may be lost. This deserves a CHANGELOG entry.

**7. gRPC exceptions not wrapped in `GrpcDurableEntityClient.java`**
Existing `DurableTaskGrpcClient` methods catch `StatusRuntimeException` and translate to domain exceptions. All new entity gRPC calls let raw gRPC exceptions escape. This breaks the `DurableEntityClient` abstraction — callers shouldn't need to catch gRPC types.

**8. Excessive INFO-level logging will flood production logs (`TaskOrchestrationExecutor.java`)**
Multiple log statements changed from `FINE` to `INFO`: every orchestrator yield, every entity call/response, every new event processed, every buffered event. In production with hundreds of orchestrations, this will flood logs. Most should remain at `FINE`/`DEBUG`.

---

### 🟡 Medium Priority

**9. `getCommittedActions()` iterates ALL pending actions, not committed count (`TaskEntityExecutor.java`)**
The method iterates all `pendingActions` rather than bounding by `committedActionCount`. It relies on a fragile invariant that both are equal at call time. If that invariant is ever violated (e.g., future refactoring), uncommitted actions leak into the result. Bound the loop by `committedActionCount`.

**10. `resultNode.asText()` loses data for non-string JSON nodes (`TaskOrchestrationExecutor.java`)**
`JsonNode.asText()` returns `""` for object/array nodes. If the DTFx ResponseMessage `"result"` field ever contains a raw JSON object, the result is silently lost. Use `resultNode.isTextual() ? resultNode.asText() : resultNode.toString()`.

**11. Legacy `callEntity` path omits `"signal"` field**
Legacy `signalEntity` JSON explicitly sets `"signal": true`, but `callEntity` does **not** set `"signal": false`. If DTFx defaults absence to `true`, all calls would be misrouted as fire-and-forget signals.

**12. `EntityQueryPageable` — `hasNext()` triggers network calls (side-effect violation)**
Both the constructor and `hasNext()` eagerly call `fetchNextPage()`, which executes gRPC queries. The `Iterator.hasNext()` contract should be a pure observation. Defer first page fetch to first `hasNext()` call.

**13. `TypedEntityMetadata` eagerly deserializes — single bad entity poisons entire query**
If `readStateAs()` throws for one entity, `TypedEntityMetadata` construction fails, halting iteration over the entire result set. No way to skip corrupt entities.

**14. Javadoc says "excluding state" but behavior includes state (`DurableTaskClient.java`)**
The deprecated `getEntityMetadata(entityId)` doc says "excluding its state" but delegates to `getEntities().getEntityMetadata(entityId)` which defaults to `includeState = true`.

**15. `EntityRunner` silently swallows malformed instance IDs**
```java
try {
    entityName = EntityInstanceId.fromString(instanceId).getName();
} catch (Exception e) {
    entityName = instanceId; // Silent fallback
}
```
This masks the real problem and produces a confusing "No entity named '...' is registered" error later.

**16. `EntityProxy.getRawClass()` doesn't handle `WildcardType`**
`Task<? extends Foo>` silently falls back to `Object.class`, losing the upper-bound type.

**17. `EntityQueryResult.entities` list not defensively copied**
The constructor stores the list reference directly. Callers can mutate via `getEntities().add(...)`. Use `Collections.unmodifiableList`.

**18. Missing interrupt flag restoration in `close()` (`DurableTaskGrpcWorker.java`)**
When `InterruptedException` is caught, `Thread.currentThread().interrupt()` is not called, violating Java's interrupt contract.

**19. `operationInfos`/results size mismatch silently truncated (`DurableTaskGrpcWorker.java`, V2 entity path)**
If sizes differ, excess infos are quietly dropped or results have no routing info. Neither case is logged. Can cause silent data loss or mis-routed responses.

**20. Lazy `cachedEntityInstanceId` without `volatile` (`EntityMetadata.java`)**
The field is lazily initialized in `getEntityInstanceId()`. Without `volatile`, it's not safe for cross-thread visibility under the Java Memory Model.

---

### 🟠 Low Priority / Design

**21. `ITaskEntity.runAsync()` is synchronous** — The method returns `Object`, not `CompletableFuture`. The `Async` suffix is misleading per universal Java convention. Rename to `run()`.

**22. `ITaskEntity` uses `I` prefix** — Inconsistent with existing SDK naming (`DataConverter`, `TaskOrchestrationContext`).

**23. `TaskOrchestrationEntityFeature.lockEntities(EntityInstanceId...)` is needlessly abstract** — Should be a concrete convenience method delegating to the `List` overload.

**24. `getEntities()` allocates a new object on every call** — The .NET SDK caches this. Cache or document.

**25. `currentCriticalSectionId` is set but never read** — Dead state. Remove or use.

**26. `lockEntities` doesn't validate duplicate entity IDs or null elements** — Same ID twice = double-lock. Null element = cryptic NPE from `Collections.sort`.

**27. No timeout mechanism for `lockEntities`** — Unlike `callEntity` with `CallEntityOptions.getTimeout()`, locks can block forever.

**28. Worker threads all share the same name `"durabletask-worker"`** — Thread dumps become useless. Use a counter.

**29. `EntityQuery.setInstanceIdStartsWith("@")` produces degenerate prefix** — Matches ALL entities.

**30. `EntityQueryPageable.cloneQuery()` duplicates `EntityQuery.copy()` logic** — Maintenance hazard if a field is added.

**31. `TaskEntityOperation` and `TaskEntityState` constructors are public** — These are internal types that users should never construct. Make package-private.

**32. `cleanEntityStorage` in `GrpcDurableEntityClient` has no max-iteration guard** — If the sidecar has a bug returning non-null tokens forever, this is an unbounded blocking call.

---

### PR Checklist

- ⚠️ CHANGELOG.md is NOT updated (PR checklist item unchecked)
- ⚠️ Merge state is "dirty" — rebase needed
- ✅ Extensive unit tests added
- ✅ E2E tests added

---

### Questions for @bachuv

1. **Re: #1 (lock close):** Is there a reason the lock release wasn't guarded with a boolean flag? `try (AutoCloseable lock = ...)` makes double-close very likely.

2. **Re: #6 (activity model change):** Was the sync→async activity execution change intentional? This is a behavioral breaking change that should at minimum be in the CHANGELOG.

3. **Re: #8 (logging):** Were the `FINE → INFO` logging changes intentional for debugging, or should they be reverted before merge?

4. **Re: #11 (missing "signal": false):** Can you confirm DTFx extension behavior when the `"signal"` field is absent from the RequestMessage JSON?

5. **Re: EVENTSENT uncommented:** `handleEventSent()` now removes pending actions for ALL `EventSent` events, not just entity-related ones. Has this been regression-tested against existing orchestrations using `sendEvent`?

6. **Re: #27 (lock timeout):** Is lock timeout deferred to a follow-up? The .NET SDK has this capability.

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -381,6 +410,335 @@ public UUID newUUID() {
...
+
+            // Create a waiter keyed by criticalSectionId
+            CompletableTask<AutoCloseable> lockTask = new CompletableTask<>();
+            TaskRecord<AutoCloseable> record = new TaskRecord<>(lockTask, "(lock)", AutoCloseable.class);
+            Queue<TaskRecord<?>> eventQueue = this.outstandingEvents.computeIfAbsent(criticalSectionId, k -> new LinkedList<>());
+            eventQueue.add(record);
+
+            // Wrap the result so that when the lock is granted, we return an AutoCloseable
+            // that releases all locks on close()
+            return lockTask.thenApply(ignored -> (AutoCloseable) () -> {
```
YunchuWang, line 703: **HIGH: Lock release `close()` is not idempotent — can cause non-deterministic replay.**

If `close()` is called twice (common with try-with-resources + explicit close, or exception-handling double-close), this sends duplicate unlock messages and increments `sequenceNumber` again, causing non-deterministic replay failures on subsequent executions. **Fix: Add a boolean guard.**

client/src/main/java/com/microsoft/durabletask/TaskOrchestrationExecutor.java
```diff
@@ -659,16 +1024,294 @@ private void handleEventRaised(HistoryEvent e) {
...
+            }
+
+            TaskRecord<?> record = outstandingQueue.remove();
+            if (outstandingQueue.isEmpty()) {
+                this.outstandingEvents.remove(criticalSectionId);
+            }
+
+            this.isInCriticalSection = true;
+            this.currentCriticalSectionId = criticalSectionId;
+            this.lockedEntityIds = this.pendingLockSets.remove(criticalSectionId);
```
YunchuWang, line 1316: **MEDIUM: `lockedEntityIds` can be null here.** If `pendingLockSets.remove()` returns null (key missing), `lockedEntityIds` is null while `isInCriticalSection` is true. This short-circuits the safety validation in `callEntity()`, allowing calls to ANY entity inside a critical section — defeating deadlock prevention. **Fix: Throw if `pendingLockSets.remove()` returns null.**

client/src/main/java/com/microsoft/durabletask/TaskEntityState.java
```diff
@@ -0,0 +1,115 @@
...
+     * Sets the entity state. The state will be serialized using the configured {@link DataConverter}.
+     *
+     * @param state the state value to set
+     */
+    public void setState(@Nullable Object state) {
+        if (state == null) {
+            deleteState();
+        } else {
+            this.serializedState = this.dataConverter.serialize(state);
+            if (this.serializedState == null) {
```
YunchuWang, line 77: **HIGH: Silent data loss.** Since the outer `if` already guarantees `state != null`, `serialize()` returning null here means a broken `DataConverter` implementation. Silently calling `deleteState()` destroys the entity's state with no error. **Fix: Throw `IllegalStateException`.**

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -340,42 +371,142 @@ public void startAndBlock() {
...
+                            }
 
-                        if (failureDetails != null) {
-                            responseBuilder.setFailureDetails(failureDetails);
-                        }
+                            if (failureDetails != null) {
+                                responseBuilder.setFailureDetails(failureDetails);
+                            }
+
+                            this.sidecarClient.completeActivityTask(responseBuilder.build());
```
YunchuWang: **CRITICAL: Activity completion failures are silently lost.** If this gRPC call throws (network issue, sidecar crash), the exception hits only the thread pool's uncaught exception handler. The activity result is permanently lost with no retry or abandon. Entity work items have an abandon fallback (see below), but activities do not. **Fix: Wrap in try-catch with abandon.**

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
YunchuWang: **HIGH: Unbounded thread pool.** `Executors.newCachedThreadPool()` has no upper bound. Activities AND entities both use this pool. A burst of work items can create unbounded threads, risking OOM. **Fix: Use a bounded `ThreadPoolExecutor` or separate pools for activities vs. entities.**

bachuv:
First set of feedback:

Critical (must fix):
1. Class.getPackageName() breaks Java 8 compatibility (project targets Java 8) - Fixed
2. maxConcurrentEntityWorkItems is accepted but never enforced -  Skipped. The .NET SDK does the same thing with no client-side enforcement.
3. Activity execution silently changed from synchronous to unbounded-concurrent - Fixed

Medium:
4. EntityMetadata.includesState incorrectly derived from proto field presence - Fixed
5. EntityQueryPageable.cloneQuery() duplicates EntityQuery.copy() and double-normalizes - Fixed
6. TaskEntityOperation constructor should be package-private - Fixed

Low / Code Quality:
7. INFO-level logging on orchestration replay hot path will flood production logs - Fixed
8. Inconsistent indentation in ENTITYREQUESTV2 handler and findMethodUncached - Fixed
9. Field declared after methods that reference it in CleanEntityStorageRequest - Fixed
10. Missing newline at end of several files - Fixed

Second set of feedback:

🔴 Critical / High Priority
1. Lock release close() is not idempotent - Fixed
2. Critical section enforcement bypassed when lockedEntityIds is null (TaskOrchestrationExecutor.java)- Fixed
3. Activity completion failures silently lost (DurableTaskGrpcWorker.java) - No longer applicable. Activities were reverted to synchronous execution (matching main). They run inline on the dispatch thread so no thread pool, no swallowed exceptions. If completeActivityTask fails, the exception propagates naturally up the call stack, same as before this PR.
4. setState() silently deletes entity state on serialization failure (TaskEntityState.java) - Fixed
5. Unbounded thread pool (DurableTaskGrpcWorker.java)
Executors.newCachedThreadPool() has no upper bound. - Fixed
6. Activity execution model silently changed from sync to async - Fixed
7. gRPC exceptions not wrapped in GrpcDurableEntityClient.java - Opened GitHub issue #274 as this was a preexisting issue. Will address in a separate PR.
8. Excessive INFO-level logging will flood production logs (TaskOrchestrationExecutor.java) - Fixed

🟡 Medium Priority
9. getCommittedActions() iterates ALL pending actions, not committed count (TaskEntityExecutor.java) - Fixed
10. resultNode.asText() loses data for non-string JSON nodes (TaskOrchestrationExecutor.java) - Fixed
11. Legacy callEntity path omits "signal" field - Fixed
12. EntityQueryPageable — hasNext() triggers network calls (side-effect violation) - Skipped, matches .NET behavior
13. TypedEntityMetadata eagerly deserializes — single bad entity poisons entire query - Skipped, matches .NET behavior
14. Javadoc says "excluding state" but behavior includes state (DurableTaskClient.java) - Fixed
15. EntityRunner silently swallows malformed instance IDs - fixed
16. EntityProxy.getRawClass() doesn't handle WildcardType - Fixed
17. EntityQueryResult.entities list not defensively copied - Fixed
18. Missing interrupt flag restoration in close() (DurableTaskGrpcWorker.java) - Fixed
19. operationInfos/results size mismatch silently truncated (DurableTaskGrpcWorker.java, V2 entity path) - Fixed
20. Lazy cachedEntityInstanceId without volatile (EntityMetadata.java) - Fixed

🟠 Low Priority / Design
21. ITaskEntity.runAsync() is synchronous — The method returns Object, not CompletableFuture. The Async suffix is misleading per universal Java convention. Rename to run(). - Fixed
22. ITaskEntity uses I prefix — Inconsistent with existing SDK naming (DataConverter, TaskOrchestrationContext) - Fixed
23. TaskOrchestrationEntityFeature.lockEntities(EntityInstanceId...) is needlessly abstract — Should be a concrete convenience method delegating to the List overload. - Fixed
24. getEntities() allocates a new object on every call — The .NET SDK caches this. Cache or document. - Fixed
25. currentCriticalSectionId is set but never read — Dead state. Remove or use. - Fixed
26. lockEntities doesn't validate duplicate entity IDs or null elements — Same ID twice = double-lock. Null element = cryptic NPE from Collections.sort. - Fixed
27. No timeout mechanism for lockEntities — Unlike callEntity with CallEntityOptions.getTimeout(), locks can block forever. - Skipped, .NET SDK also has no lock timeout
28. Worker threads all share the same name "durabletask-worker" — Thread dumps become useless. Use a counter. - Fixed
29. EntityQuery.setInstanceIdStartsWith("@") produces degenerate prefix — Matches ALL entities. - Skipped. matches .NET SDK behavior
30. EntityQueryPageable.cloneQuery() duplicates EntityQuery.copy() logic — Maintenance hazard if a field is added. - Fixed
31. TaskEntityOperation and TaskEntityState constructors are public — These are internal types that users should never construct. Make package-private. - Fixed
32. cleanEntityStorage in GrpcDurableEntityClient has no max-iteration guard — If the sidecar has a bug returning non-null tokens forever, this is an unbounded blocking call. - Skipped, matches .NET SDK behavior

Questions

Re: #1 (lock close): This was an oversight. The boolean guard has now been added. Double-close is idempotent: first call emits unlocks, subsequent calls are no-ops.

Re: #6 (activity model change): This has been reverted so activities are back to synchronous execution on the dispatch thread, identical to main. No behavioral change, no CHANGELOG entry needed. The thread pool is now used exclusively for entity work items.

Re: #8 (logging): The FINE → INFO changes were not intentional for production use. They were debug-time changes that should have been reverted. All hot-path logs have been restored to FINE. Only entity operation failures remain at INFO since those are genuinely noteworthy.

Re: #11 (missing "signal": false): DTFx RequestMessage.IsSignal is a C# bool property with [DataMember(Name = "signal", EmitDefaultValue = false)]. It defaults to false, and EmitDefaultValue = false means the .NET serializer omits the field when the value is false. So absent and false are functionally identical. DTFx treats both as "this is a call, not a signal." We now explicitly set "signal": false for clarity.

Re: EVENTSENT uncommented: Could you point to the specific lines? I want to verify the exact change before answering. The handleEventSent handler should only process entity-related events. If it's broader than that, it needs to be scoped.

Re: #27 (lock timeout): Yes, deferred. The .NET SDK's LockEntitiesAsync does not have a timeout parameter. It returns Task<IAsyncDisposable> with no timeout overload. This would be a new feature for both SDKs, not a parity gap.

YunchuWang:
## Review Finding: Missing `continueAsNew` guard inside critical sections

`continueAsNew()` in `TaskOrchestrationExecutor.ContextImplTask` does not check `isInCriticalSection`. This allows users to call `continueAsNew()` while holding entity locks.

**Why this is a problem:**

When `continueAsNew()` is called inside a `try (AutoCloseable lock = ...)` block:

1. `ContinueAsNewInterruption` is thrown
2. Java try-with-resources calls `close()` - unlock actions are added to `pendingActions`
3. The interruption propagates to `execute()` - `CONTINUED_AS_NEW` completion action is added
4. All actions are sent to the sidecar in a single response

However, the sidecar's handling of unlock + continueAsNew actions in the same response is **underspecified** - we don't know whether unlocks are processed before the restart, or whether this behaves consistently across sidecar implementations (DTS, DTFx, Dapr).

Even if unlocks are processed first, there is a **race window** between the lock release and the new instance re-acquiring locks, during which other orchestrations can modify the locked entities - breaking the atomicity that critical sections guarantee.

**The .NET SDK blocks this:**
```csharp
if (this.IsInsideCriticalSection)
    throw new InvalidOperationException("Cannot continue-as-new while in a critical section.");
```

**Suggested fix:**
```java
@Override
public void continueAsNew(Object input, boolean preserveUnprocessedEvents) {
    Helpers.throwIfOrchestratorComplete(this.isComplete);
    if (this.isInCriticalSection) {
        throw new IllegalStateException(
            "Cannot continue-as-new while inside a critical section. "
            + "Exit the critical section first by closing the lock.");
    }
    // ... existing code
}
```

**Correct user pattern:**
```java
try (AutoCloseable lock = ctx.getEntities().lockEntities(e1, e2).await()) {
    // transactional work
}
// After exiting critical section:
ctx.continueAsNew(newInput);
```

YunchuWang (dismissed):
## ✅ Approved — Durable Entity Parity Review

> **Disclosure**: This review was generated by AI (GitHub Copilot CLI, Claude Opus 4.6) and approved/granted by reviewer @YunchuWang. The AI performed a systematic cross-SDK parity analysis reading all 77 changed files in this PR against the .NET reference implementation in [microsoft/durabletask-dotnet](https://github.com/microsoft/durabletask-dotnet). The reviewer verified the findings and approves this PR.

---

### Verdict: **Mostly aligned with minor gaps — safe to merge**

This PR delivers a comprehensive Durable Entity implementation for the Java SDK that covers all major feature categories present in the .NET reference. The behavioral model (transactional batch execution, per-operation commit/rollback of state AND actions, signal vs call semantics, entity locking/critical sections) is correctly implemented and well-tested.

**Strengths:**
- ✅ Core entity model (interface + abstract base + reflection dispatch) fully aligned with .NET
- ✅ All .NET entity client capabilities present (signal, query, metadata, cleanup)
- ✅ Azure Functions integration complete (trigger, middleware, client context)
- ✅ Entity locking with `AutoCloseable` correctly maps .NET's `IAsyncDisposable`
- ✅ Test suite (182+ tests) **significantly exceeds** .NET reference test coverage (~47 abstraction tests)
- ✅ 5 E2E integration tests validate real workflows
- ✅ Non-determinism detection tested for signal/call/lock operations
- ✅ Useful Java-only additions: `CallEntityOptions.timeout`, `AbstractTaskEntity.dispatch()` re-entrant calls

### Suggested Follow-Up Issues

The following items are minor and should **not block merge**, but would improve parity and robustness:

---

**1. `EntityMetadata.readStateAs()` should guard when state not included** (Medium)

.NET's `EntityMetadata<T>.State` throws when `IncludesState=false`, protecting against silent bugs where callers forget to request state inclusion. Java's `readStateAs()` silently returns `null`, making "state doesn't exist" indistinguishable from "state wasn't fetched". Suggest throwing `IllegalStateException` when `includesState == false`.

**2. Verify distributed trace context propagation in `GrpcDurableEntityClient`** (Medium)

.NET's `GrpcDurableEntityClient.SignalEntityAsync` explicitly captures `Activity.Current` and propagates trace context in gRPC calls. Please verify Java's implementation does the same for observability parity.

**3. Verify lock ordering in `lockEntities()` to prevent deadlocks** (Medium)

.NET sorts entity IDs before acquiring locks to prevent deadlocks when multiple orchestrations lock overlapping entity sets. Verify the Java implementation applies the same sorting.

**4. Add `getState(Class<T>, T defaultValue)` overload to `TaskEntityState`** (Low)

.NET offers `GetState<T>(T? defaultValue = default)` for convenience. Java's `getState(Class<T>)` always returns null when no state exists. Adding a defaultValue overload improves ergonomics.

**5. Document Java-specific additions for cross-SDK awareness** (Low)

`CallEntityOptions.timeout` (with timer-based implementation) and `AbstractTaskEntity.dispatch()` (re-entrant self-calls) are useful additions not present in .NET. These should be documented as Java-specific features for future cross-SDK alignment discussions.

**6. Add `EnableEntitySupport` configuration option** (Low)

.NET's `DurableTaskClientOptions.EnableEntitySupport` allows disabling entity client creation. Adding a similar toggle to Java would provide configuration parity for environments that don't use entities.

---

### Language-Justified Differences (NOT issues)

These differences are expected and correct:
- `ValueTask<object?>` → synchronous `Object` return (Java entity execution model)
- `CancellationToken` → not present (Java uses thread interruption)
- `InCriticalSection(out list)` → `isInCriticalSection()` + `getLockedEntities()` (no `out` params in Java)
- Source generators → JDK dynamic proxy for typed entity proxies
- Immutable records → mutable builder-style options classes
- `IAsyncDisposable` → `AutoCloseable` for lock release

LGTM 🚀

YunchuWang (approved):

PR #272, opened 2026-03-23
Add large payload support

bachuv:

Adding [large payload support ](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-task-scheduler/durable-task-scheduler-large-payloads?tabs=csharp&pivots=durable-functions)in the Java SDK.

## Add large payload support for Durable Task Java SDK

Adds the ability to externalize[ large payloads](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-task-scheduler/durable-task-scheduler-large-payloads?tabs=csharp&pivots=durable-functions) (900 KB – 10 MB) to Azure Blob Storage, preventing gRPC message size failures (~4 MB limit). Payloads below 900 KB continue inline with no behavior change. This feature is fully supported for standalone DTS (Durable Task Scheduler) applications. Azure Functions support requires a corresponding .NET extension update and will be added in another PR.

### Usage (standalone DTS)

```java
// 1. Configure blob storage for payload externalization
BlobPayloadStoreOptions storeOptions = new BlobPayloadStoreOptions.Builder()
    .setConnectionString("DefaultEndpointsProtocol=https;AccountName=...")
    .build();

// 2. Create worker with large payload support
DurableTaskGrpcWorkerBuilder workerBuilder =
    DurableTaskSchedulerWorkerExtensions.createWorkerBuilder(dtsConnectionString);
AzureBlobPayloadsExtensions.useBlobStoragePayloads(workerBuilder, storeOptions);
workerBuilder.addOrchestration(...);
workerBuilder.addActivity(...);
DurableTaskGrpcWorker worker = workerBuilder.build();

// 3. Create client with large payload support
DurableTaskGrpcClientBuilder clientBuilder =
    DurableTaskSchedulerClientExtensions.createClientBuilder(dtsConnectionString);
AzureBlobPayloadsExtensions.useBlobStoragePayloads(clientBuilder, storeOptions);
DurableTaskClient client = clientBuilder.build();

// payloads >= 900 KB are automatically externalized to blob storage
// and resolved transparently on the receiving side.
```

### How it works

- **Outbound**: When a payload (activity input, orchestration output, etc.) exceeds 900 KB, it is gzip-compressed, uploaded to Azure Blob Storage, and replaced with a small token (`blob:v1:<container>:<blobName>`)
- **Inbound**: When the SDK encounters a token in history events or activity inputs, it downloads and decompresses the original payload transparently
- **Payloads > 10 MB** are rejected with `PayloadTooLargeException`
- **Thresholds match the .NET SDK** (900 KB default) for cross-SDK compatibility

### Code breakdown

**New module: `azure-blob-payloads`**
- `BlobPayloadStore` — `PayloadStore` implementation that uploads/downloads payloads to Azure Blob Storage with gzip compression
- `BlobPayloadStoreOptions` — Builder-pattern configuration (connection string, endpoint+credential, or pre-configured `BlobServiceClient`; container name, blob prefix, compression toggle)
- `BlobPayloadStoreProvider` — `PayloadStoreProvider` SPI implementation for auto-discovery via `ServiceLoader` when `DURABLETASK_LARGE_PAYLOADS_CONNECTION_STRING` env var is set
- `AzureBlobPayloadsExtensions` — Convenience methods to configure blob storage payloads on worker/client builders

**Changes to `client` module**
- `PayloadStore` — Interface for upload/download/token detection (extensible to non-blob implementations)
- `PayloadStoreProvider` — SPI interface for `ServiceLoader`-based auto-discovery
- `PayloadHelper` — Internal threshold-checking and store-delegation logic (null/empty guard, fast char-length check, UTF-8 byte-length check, max-cap rejection, upload)
- `PayloadInterceptionHelper` — Resolves tokens in all protobuf history event types (`ExecutionStarted`, `TaskCompleted`, `TaskScheduled`, `SubOrchestrationInstanceCreated`, `EventRaised`, etc.) and externalizes payloads in orchestrator response actions (`ScheduleTaskAction`, `CreateSubOrchestrationAction`, `CompleteOrchestrationAction`)
- `LargePayloadOptions` — Configurable thresholds (default 900 KB externalization, 10 MB max)
- `PayloadTooLargeException` — Thrown when payload exceeds max
- `DurableTaskGrpcWorkerBuilder.useExternalizedPayloads()` — Configures worker with `PayloadStore`
- `DurableTaskGrpcClientBuilder.useExternalizedPayloads()` — Configures client with `PayloadStore`
- `DurableTaskGrpcWorker` — Resolves activity request payloads, externalizes activity response payloads
- `DurableTaskGrpcClient` — Externalizes orchestration inputs/events, resolves instance metadata outputs
- `OrchestrationRunner` — Resolves incoming history event payloads, externalizes outgoing action payloads

**Changes to `azurefunctions` module**
- `OrchestrationMiddleware` — Extended to intercept `DurableActivityTrigger` (resolves input tokens, externalizes large outputs) and uses a resolve-only `PayloadStore` wrapper for orchestrations (resolves history tokens but does not externalize action inputs, since the .NET extension dispatches those)

**Tests**
- `PayloadHelperTest` — Threshold logic, null/empty handling, UTF-8 byte counting
- `PayloadInterceptionHelperTest` — Protobuf history event resolution, action externalization
- `LargePayloadOptionsTest` — Builder validation
- `OrchestrationRunnerPayloadTest` — End-to-end `OrchestrationRunner` with `PayloadStore`
- `DurableTaskGrpcWorkerBuilderTest` — Builder configuration validation
- `LargePayloadIntegrationTests` — Integration tests against DTS emulator + Azurite
- `BlobPayloadStoreOptionsTest` — Options builder validation

### Pull request checklist

### Additional information

Additional PR information


azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
```diff
@@ -25,6 +31,23 @@
...
+
+    /**
+     * Environment variable for the Azure Storage connection string used for large payload externalization.
+     */
+    private static final String ENV_STORAGE_CONNECTION_STRING = "DURABLETASK_STORAGE_CONNECTION_STRING";
+
+    /**
+     * Fallback environment variable (standard Azure Functions storage connection).
+     */
+    private static final String ENV_AZURE_WEB_JOBS_STORAGE = "AzureWebJobsStorage";
```
YunchuWang: **Bug (Critical): Falling back to `AzureWebJobsStorage` silently enables large-payload externalization for ALL Azure Functions apps.**

`AzureWebJobsStorage` is a **required** setting in virtually every Azure Functions app. This means large-payload externalization will be silently enabled for everyone who upgrades, even if they don't want it. This introduces:
- Extra latency on every payload (blob upload/download calls for threshold checks)
- Unexpected Azure Storage costs (a new `durabletask-payloads` container created automatically)
- Potential failures if storage permissions are insufficient for container creation

This should require **explicit opt-in**, not an implicit fallback. Only use `DURABLETASK_STORAGE_CONNECTION_STRING` or a similar dedicated env var. Remove the `AzureWebJobsStorage` fallback.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/OrchestrationMiddleware.java
```diff
@@ -12,10 +12,16 @@
 import com.microsoft.durabletask.CompositeTaskFailedException;
 import com.microsoft.durabletask.DataConverter;
 import com.microsoft.durabletask.OrchestrationRunner;
+import com.microsoft.durabletask.PayloadStore;
 import com.microsoft.durabletask.TaskFailedException;
+import com.microsoft.durabletask.azureblobpayloads.BlobPayloadStore;
```
YunchuWang: **Design (Critical): Hard dependency on `BlobPayloadStore` in middleware layer.**

The `OrchestrationMiddleware` (in the `azurefunctions` module) directly imports and instantiates `BlobPayloadStore` from `azure-blob-payloads`. This:
1. Forces the Azure Functions binding to always pull in Azure Blob Storage SDK dependencies
2. Prevents users from using a custom `PayloadStore` implementation
3. Violates the `PayloadStore` abstraction you carefully designed

Consider using SPI (ServiceLoader) to discover `PayloadStore` implementations, or accept it via configuration/environment, rather than hardcoding `BlobPayloadStore`.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
```diff
@@ -0,0 +1,198 @@
...
+        String blobName = this.blobPrefix + UUID.randomUUID().toString().replace("-", "") + BLOB_EXTENSION;
+        BlobClient blobClient = this.containerClient.getBlobClient(blobName);
+
+        byte[] rawData = payload.getBytes(StandardCharsets.UTF_8);
+        byte[] data;
+
+        if (this.compressPayloads) {
+            data = gzipCompress(rawData);
+            blobClient.upload(new ByteArrayInputStream(data), data.length, true);
+            blobClient.setHttpHeaders(new BlobHttpHeaders().setContentEncoding(GZIP_CONTENT_ENCODING));
```
YunchuWang: **Performance (Medium): Two HTTP calls for compressed uploads.**

`upload()` + `setHttpHeaders()` = 2 network round-trips. Set content encoding at upload time instead:

```java
BlobHttpHeaders headers = new BlobHttpHeaders().setContentEncoding(GZIP_CONTENT_ENCODING);
BlobParallelUploadOptions uploadOptions = new BlobParallelUploadOptions(
    new ByteArrayInputStream(data), data.length)
    .setHeaders(headers);
blobClient.uploadWithResponse(uploadOptions, null, null);
```

This halves the latency for every compressed upload.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
```diff
@@ -0,0 +1,198 @@
...
+        String blobName = extractBlobName(token);
+        BlobClient blobClient = this.containerClient.getBlobClient(blobName);
+
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        blobClient.downloadStream(outputStream);
+
+        byte[] rawBytes = outputStream.toByteArray();
+
+        // Check if the blob was compressed by inspecting content encoding
+        String contentEncoding = blobClient.getProperties().getContentEncoding();
```
YunchuWang: **Performance (Medium): Extra HTTP call to check content encoding on every download.**

`downloadStream()` + `getProperties()` = 2 network round-trips for every download. Instead, use `downloadContent()` which returns a `BlobDownloadContentResponse` containing both data and properties:

```java
BlobDownloadContentResponse response = blobClient.downloadContentWithResponse(null, null, null, null);
byte[] rawBytes = response.getValue().toBytes();
String contentEncoding = response.getDeserializedHeaders().getContentEncoding();
```

Alternatively, encode compression info in the token format (e.g., `blob:v1:gz:<container>:<blobName>`) to avoid needing to query properties at all.

client/src/main/java/com/microsoft/durabletask/PayloadHelper.java
```diff
@@ -0,0 +1,93 @@
...
+     * @return the original value if below threshold, or an opaque token if externalized
+     * @throws IllegalArgumentException if the payload exceeds the maximum externalized payload size
+     */
+    String maybeExternalize(String value) {
+        // (1) null/empty guard
+        if (value == null || value.isEmpty()) {
+            return value;
+        }
+
+        int byteSize = value.getBytes(StandardCharsets.UTF_8).length;
```
YunchuWang: **Performance (Medium): Full UTF-8 byte array allocation just to check size.**

`value.getBytes(StandardCharsets.UTF_8).length` creates a complete copy of the string as bytes solely to measure its size. For a 10MB string, this temporarily doubles memory usage.

Fast-path optimization: since `value.length() <= UTF-8 byte length` always holds (each Java char encodes to 1-3 UTF-8 bytes), check the char count first:

```java
// Fast path: if char count is below threshold, byte count is too
if (value.length() <= this.options.getThresholdBytes()) {
    return value;
}
// Only allocate bytes if char count exceeds threshold
int byteSize = value.getBytes(StandardCharsets.UTF_8).length;
```

This avoids the allocation entirely for the common case (small payloads).

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
```diff
@@ -0,0 +1,198 @@
...
+        return token;
+    }
+
+    @Override
+    public String download(String token) {
+        if (token == null || token.isEmpty()) {
+            throw new IllegalArgumentException("token must not be null or empty");
+        }
+
+        String blobName = extractBlobName(token);
```
YunchuWang, line 118: **Bug (Low): `extractBlobName()` ignores the container name embedded in the token.**

The token format is `blob:v1:<container>:<blobName>`, but this method only extracts the blob name and discards the container. If a token was created by a `BlobPayloadStore` pointing to a different container (e.g., in a migration scenario or multi-tenant setup), this will attempt to download from the wrong container.

At minimum, validate that the container in the token matches `this.containerName` and throw if not. Ideally, use the container from the token for the download.

client/src/main/java/com/microsoft/durabletask/PayloadStore.java
```diff
@@ -0,0 +1,47 @@
...
+ * @see LargePayloadOptions
+ */
+public interface PayloadStore {
+    /**
+     * Uploads a payload string to external storage and returns an opaque token reference.
+     *
+     * @param payload the payload data to upload
+     * @return an opaque token string that can be used to retrieve the payload via {@link #download(String)}
+     */
+    String upload(String payload);
```
YunchuWang, line 37: **Design (Low): No `delete()` method — externalized payloads accumulate forever.**

The `PayloadStore` interface has `upload()` and `download()` but no mechanism for cleanup. Externalized blobs will accumulate indefinitely in storage, increasing costs over time.

Consider:
1. Adding a `delete(String token)` method to the interface
2. Documenting that users should configure Azure Blob Storage lifecycle management policies
3. Or both — the framework could call `delete()` when orchestrations complete/purge

At minimum, add documentation about the retention/cleanup story.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
```diff
@@ -0,0 +1,198 @@
...
+                .credential(options.getCredential())
+                .buildClient();
+        }
+
+        this.containerClient = serviceClient.getBlobContainerClient(options.getContainerName());
+        this.blobPrefix = options.getBlobPrefix();
+        this.containerName = options.getContainerName();
+        this.compressPayloads = options.isCompressPayloads();
+
+        ensureContainerExists();
```
YunchuWang: **Design (Low): Constructor makes a network call (`ensureContainerExists`).**

Creating a `BlobPayloadStore` instance immediately calls `ensureContainerExists()` which makes HTTP requests to Azure Storage. If the storage account is temporarily unavailable, the entire application fails to start.

Consider lazy initialization — check/create the container on the first `upload()` call instead. This makes the constructor pure and the failure point more predictable.

client/src/main/java/com/microsoft/durabletask/PayloadInterceptionHelper.java
```diff
@@ -0,0 +1,326 @@
...
+
+    @FunctionalInterface
+    interface HistoryEventUpdater {
+        HistoryEvent apply(HistoryEvent event, StringValue newValue);
+    }
+
+    private static HistoryEvent resolveStringValueField(HistoryEvent event, StringValue field,
+                                                         PayloadHelper payloadHelper,
+                                                         HistoryEventUpdater updater) {
+        if (!field.isInitialized() || field.getValue().isEmpty()) {
```
YunchuWang: **Code Quality: `field.isInitialized()` is always `true` in protobuf 3.**

In proto3, all fields are implicitly optional and `isInitialized()` always returns `true`. This check is a no-op and could be misleading to future readers. The `field.getValue().isEmpty()` check on the same line is the one that actually gates the logic.

Remove the `!field.isInitialized()` check for clarity, or replace with a comment explaining the intent.

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -289,7 +304,33 @@ public void startAndBlock() {
...
                                     .build();
 
-                            this.sidecarClient.completeOrchestratorTask(response);
+                            // Externalize large payloads and send (with optional chunking).
+                            // If externalization or chunking fails, report as orchestration failure.
+                            try {
+                                if (this.payloadHelper != null) {
+                                    response = externalizeOrchestratorResponsePayloads(response);
+                                }
+                                sendOrchestratorResponse(response);
```
YunchuWang, line 313: **Robustness: Externalization failure during response send converts to orchestration failure.**

The try-catch around `externalizeOrchestratorResponsePayloads` + `sendOrchestratorResponse` catches `IllegalArgumentException | IllegalStateException` and reports the orchestration as FAILED. This is reasonable for max-payload-exceeded errors, but it will also catch programming bugs (e.g., a null pointer wrapped in IllegalStateException from the payload store). Consider:
1. Using a more specific exception type for payload-too-large errors
2. Logging the full exception including the stack trace (currently only `e.getMessage()` is in the failure details)
3. The `StackTrace` field of `TaskFailureDetails` is not being set here, unlike other failure paths

YunchuWang (changes requested):
## Code Review: Large Payload Support

Good overall design — the `PayloadStore` abstraction, `PayloadHelper` threshold logic, and `PayloadInterceptionHelper` protobuf interception are well-structured. The response chunking implementation is solid. However, I found several issues that need attention.

### Summary of Findings

**Critical:**
1. `OrchestrationMiddleware` silently enables blob storage for ALL Azure Functions apps (falls back to `AzureWebJobsStorage`)
2. `OrchestrationMiddleware` hardcodes `BlobPayloadStore` — violates the `PayloadStore` abstraction

**Medium (Performance):**
3. `BlobPayloadStore.upload()` makes 2 HTTP calls for compressed uploads (upload + setHttpHeaders)
4. `BlobPayloadStore.download()` makes 2 HTTP calls (downloadStream + getProperties)
5. `PayloadHelper.maybeExternalize()` allocates full UTF-8 byte array just to check size

**Low / Design:**
6. `BlobPayloadStore.extractBlobName()` ignores container name from token
7. `PayloadStore` interface has no `delete()` — externalized payloads accumulate forever
8. `PayloadInterceptionHelper` — `field.isInitialized()` is always true in proto3
9. `BlobPayloadStore` constructor makes network call (ensureContainerExists)

YunchuWang (approved):
LGTM. Thorough review of all 27 changed files. Well-architected large payload support with clean interface abstraction (PayloadStore/PayloadStoreProvider), comprehensive protobuf interception at all 4 boundary points (client, worker, orchestration runner, AzFunctions middleware), correct chunking algorithm, and strong test coverage (1859 lines including 853-line integration suite).

Key strengths:
- Clean separation: client module has zero blob storage deps; azure-blob-payloads provides one implementation via SPI
- Security: decompression bomb protection (20 MiB cap), container name validation preventing cross-container access, no SAS tokens in payload tokens
- Backward compatible: zero overhead when externalization is not configured
- Correct JDK 8 compat in production code

Minor non-blocking suggestions for follow-up:
1. Consider adding TaskFailedEvent.failureDetails interception in resolveHistoryEvent for cross-SDK completeness
2. Consider adding Content-Type: application/json header on blob uploads for better Azure Portal observability
3. BlobPayloadStoreOptionsTest.multipleAuthMethods_throws test name is slightly misleading (actually tests endpoint-without-credential)

None of these are blockers.

PR #280, opened 2026-04-13
Adding large payload support for the standalone SDK

bachuv:
Adding large payload support for the standalone SDK

### Pull request checklist


azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
```diff
@@ -0,0 +1,779 @@
...
+    /**
+     * Computes the UTF-8 encoded byte length of a string.
+     * <p>
+     * Uses Java's canonical UTF-8 encoding behavior so malformed surrogate
+     * sequences are measured exactly the same way as payload serialization.
+     */
+    private static int utf8ByteLength(String s) {
+        return s.getBytes(StandardCharsets.UTF_8).length;
+    }
+}
```
YunchuWang, line 1291: **[F1] High / P1 — `utf8ByteLength` allocates a large byte array on every call**

`s.getBytes(StandardCharsets.UTF_8).length` allocates a full copy of the string as a byte array just to measure its length. `maybeExternalize` is called on every `StringValue` field in every gRPC message — an `OrchestratorResponse` with multiple actions can trigger this many times. For strings near the 900KB threshold, each call allocates ~900KB+ of garbage.

**Impact**: Significant GC pressure under high-throughput scenarios (many concurrent orchestrations returning large payloads).

**Suggestion**: Compute UTF-8 length by iterating chars — zero allocation:
```java
private static int utf8ByteLength(String s) {
    int count = 0;
    for (int i = 0; i < s.length(); i++) {
        char c = s.charAt(i);
        if (c <= 0x7F) count++;
        else if (c <= 0x7FF) count += 2;
        else if (Character.isHighSurrogate(c)) { count += 4; i++; }
        else count += 3;
    }
    return count;
}
```

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -566,12 +577,162 @@ private GetWorkItemsRequest buildGetWorkItemsRequest() {
...
+            if (hasTraceContext) {
+                estimate.setOrchestrationTraceContext(traceContext);
+            }
+        } else {
+            estimate.setNumEventsProcessed(Int32Value.of(0));
+        }
+
+        return estimate.build().getSerializedSize();
+    }
+
```
YunchuWang, line 1000: **[F2] High / P1 — `estimateChunkSerializedSize` clones and builds the entire protobuf on every iteration**

The inner loop calls `chunk.clone() ... .build().getSerializedSize()` for each action added. This is O(N²) — each clone copies all previously-accumulated actions. For orchestrations with many actions (e.g., 100+ `callActivity` fan-outs), this creates significant latency.

**Suggestion**: Cache the envelope overhead size (computed once), then incrementally track size by adding each new action's `getSerializedSize()` + ~5 bytes varint overhead. Only do the expensive clone+build when the estimate is close to the limit (within 1% margin).

YunchuWang (commented):
## Code Review Summary

**Change intent**: Add large payload externalization support to the standalone Durable Task Java SDK via a new `azure-blob-payloads` module. Payloads exceeding a configurable threshold are transparently uploaded to Azure Blob Storage through a gRPC `ClientInterceptor`, with automatic orchestrator response chunking for oversized gRPC messages.

**Overall risk**: Moderate
**Merge recommendation**: Safe with fixes — two performance issues (F1, F2) should be addressed before merge.

### Architecture

This is a well-designed refactoring that moves payload externalization from application-level (manual per-call-site handling) down to the transport layer (gRPC interceptor). Key strengths:
- **Transparency** — application code (orchestrations, activities) is completely unaware of large payloads
- **Separation of concerns** — externalization logic concentrated in one interceptor
- **Extensibility** — `addInterceptor` API is generic, future interceptors (encryption, compression) can be added
- **Cross-SDK alignment** — chunking logic matches .NET SDK behavior
- **Test quality** — comprehensive coverage with mock unit tests and Azurite-backed integration tests

### Findings by Severity
- High: 2
- Medium: 3
- Low: 2
- Nit: 2

**Proto changes note** (N1): The proto file includes unrelated additions (`ActivityRequest.tags`, `RewindOrchestrationAction`, `PurgeInstanceFilter.timeout`). These appear to be upstream proto sync. No functional impact, but worth noting in the PR description if intentional.

**Sample note** (N2): `LargePayloadSample.java` line 123 hardcodes the `blob:v1:` token prefix check as a defensive assertion. Acceptable for a sample, but will need updating if the token format ever changes.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
```diff
@@ -0,0 +1,228 @@
...
+        // Ensure container exists (idempotent)
+        try {
+            this.containerClient.createIfNotExists();
+        } catch (BlobStorageException e) {
+            // 409 Conflict means it already exists — safe to ignore
+            if (e.getStatusCode() != 409) {
+                throw new PayloadStorageException(
+                    "Failed to create blob container '" + this.containerClient.getBlobContainerName() + "'.", e);
+            }
+        }
```
YunchuWang: **[F3] Medium / P2 — `createIfNotExists` called on every upload**

In steady state the container already exists, but every `upload()` call makes an extra HTTP request to check. Under high-frequency upload scenarios (multiple activities returning large payloads concurrently), this adds unnecessary latency and Azure Storage API calls.

**Suggestion**: Use an `AtomicBoolean` to skip after first confirmation:
```java
private final AtomicBoolean containerVerified = new AtomicBoolean(false);

// In upload():
if (!containerVerified.get()) {
    containerClient.createIfNotExists();
    containerVerified.set(true);
}
```

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
```diff
@@ -0,0 +1,779 @@
...
+                                .setErrorType(ex.getClass().getName())
+                                .setErrorMessage(ex.getMessage())
+                                .setStackTrace(StringValue.of(getStackTraceString(ex)))
+                                .setIsNonRetriable(true)
+                                .build())
+                            .build())
+                        .build())
+                    .build();
+            }
+            throw ex instanceof RuntimeException ? (RuntimeException) ex : new RuntimeException(ex);
```
YunchuWang: **[F4] Medium / P2 — Non-permanent storage exceptions become RuntimeException, may leave orchestrations stuck**

When `externalizeOrchestratorResponse` catches a non-permanent exception (e.g., transient network failure), it re-throws as `RuntimeException`. This propagates through gRPC `sendMessage`, causing the call to fail. The worker's `startAndBlock` loop only retries on `StatusRuntimeException` — other `RuntimeException`s from individual work item processing are not gracefully handled. The orchestration won't receive a completion (success or failure), potentially leaving it stuck until timeout.

**Suggestion**: Consider converting transient storage failures into a retriable orchestration failure (`isNonRetriable=false`), or wrap the `completeOrchestratorTask` call in the worker loop with a try-catch that abandons the work item on unexpected exceptions.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
```diff
@@ -0,0 +1,779 @@
...
+        }
+    }
+
+    private OrchestratorAction externalizeOrchestratorAction(OrchestratorAction action) {
```
YunchuWang, line 260: **[F5] Medium / P2 — `externalizeOrchestratorAction` uses independent `if` blocks for a protobuf `oneof` field**

`OrchestratorAction.orchestratorActionType` is a protobuf `oneof` — exactly one case is active at a time. But this method checks all 6 cases with independent `if` blocks instead of `else if` or `switch`. Functionally correct (only one matches), but wasteful and inconsistent with `resolveEventPayloads` which already uses `switch (e.getEventTypeCase())`.

**Suggestion**: Use `switch (action.getOrchestratorActionTypeCase())` for consistency and clarity.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadStorageOptions.java
```diff
@@ -0,0 +1,205 @@
...
+     * @return this options object
+     */
+    public LargePayloadStorageOptions setMaxPayloadBytes(int maxPayloadBytes) {
+        if (maxPayloadBytes <= 0) {
+            throw new IllegalArgumentException(
+                "Maximum payload size must be a positive number of bytes.");
+        }
+        this.maxPayloadBytes = maxPayloadBytes;
+        return this;
+    }
```
YunchuWang, line 97: **[F7] Low / P3 — No validation that `maxPayloadBytes >= thresholdBytes`**

It's possible to set `maxPayloadBytes` smaller than `thresholdBytes`. In that case, any payload exceeding the threshold would also exceed the max, causing `maybeExternalize` to throw `PayloadStorageException` instead of externalizing — effectively disabling the feature with a misleading error message.

**Suggestion**: Validate `maxPayloadBytes >= thresholdBytes` in the `LargePayloadInterceptor` constructor or in the options setters.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang, line 1291: **[F1] High / P1 — `utf8ByteLength` allocates a large byte array on every call**

`s.getBytes(StandardCharsets.UTF_8).length` allocates a full copy of the string as a byte array just to measure its length. `maybeExternalize` is called on every `StringValue` field in every gRPC message — an `OrchestratorResponse` with multiple actions can trigger this many times. For strings near the 900KB threshold, each call allocates ~900KB+ of garbage.

**Impact**: Significant GC pressure under high-throughput scenarios.

**Suggestion**: Compute UTF-8 length by iterating chars (zero allocation):
```java
private static int utf8ByteLength(String s) {
    int count = 0;
    for (int i = 0; i < s.length(); i++) {
        char c = s.charAt(i);
        if (c <= 0x7F) count++;
        else if (c <= 0x7FF) count += 2;
        else if (Character.isHighSurrogate(c)) { count += 4; i++; }
        else count += 3;
    }
    return count;
}
```

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
YunchuWang, line 1000: **[F2] High / P1 — `estimateChunkSerializedSize` clones and builds entire protobuf on every iteration**

The inner loop calls `chunk.clone()...build().getSerializedSize()` for each action added. This is O(N²) — each clone copies all previously-accumulated actions. For orchestrations with many actions (e.g., 100+ `callActivity` fan-outs), this creates significant latency.

**Suggestion**: Cache the envelope overhead size (computed once), then incrementally track size by adding each new action's `getSerializedSize()` + ~5 bytes varint overhead. Only do the expensive clone+build when the running estimate is close to the limit.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
YunchuWang: **[F3] Medium / P2 — `createIfNotExists` called on every upload**

In steady state the container already exists, but every `upload()` call makes an extra HTTP round-trip to check. Under high-frequency upload scenarios (multiple activities returning large payloads concurrently), this adds unnecessary latency.

**Suggestion**: Use an `AtomicBoolean` to skip after first confirmation:
```java
private final AtomicBoolean containerVerified = new AtomicBoolean(false);

// In upload():
if (!containerVerified.get()) {
    containerClient.createIfNotExists();
    containerVerified.set(true);
}
```

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang: **[F4] Medium / P2 — Non-permanent storage exceptions become RuntimeException, may leave orchestrations stuck**

When `externalizeOrchestratorResponse` catches a non-permanent exception (e.g., transient network failure), it re-throws as `RuntimeException`. This propagates through gRPC `sendMessage`, causing the call to fail. The worker's main loop only retries on `StatusRuntimeException` — other exceptions from individual work items are not gracefully handled. The orchestration won't receive a completion notification, potentially leaving it stuck until timeout.

**Suggestion**: Consider converting transient storage failures into a retriable orchestration failure (`isNonRetriable=false`), or wrap `completeOrchestratorTask` in the worker loop with a try-catch that abandons the work item on unexpected exceptions.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang, line 260: **[F5] Medium / P2 — `externalizeOrchestratorAction` uses independent `if` blocks for a protobuf `oneof` field**

`OrchestratorAction.orchestratorActionType` is a protobuf `oneof` — exactly one case is active at a time. But this method checks all 6 cases with independent `if` blocks instead of `else if` or `switch`. Functionally correct (only one matches), but wasteful and inconsistent with `resolveEventPayloads` (line 536) which correctly uses `switch (e.getEventTypeCase())`.

**Suggestion**: Use `switch (action.getOrchestratorActionTypeCase())` for consistency.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadStorageOptions.java
YunchuWang, line 97: **[F7] Low / P3 — No validation that `maxPayloadBytes >= thresholdBytes`**

It's possible to set `maxPayloadBytes` smaller than `thresholdBytes`. In that case, any payload exceeding the threshold would also exceed the max, causing `maybeExternalize` to throw `PayloadStorageException` instead of externalizing — effectively disabling the feature with a misleading error message.

**Suggestion**: Validate `maxPayloadBytes >= thresholdBytes` in the `LargePayloadInterceptor` constructor.

YunchuWang (commented):
## Code Review Summary

**Change intent**: Add large payload externalization support to the standalone Durable Task Java SDK via a new `azure-blob-payloads` module. Payloads exceeding a configurable threshold are transparently uploaded to Azure Blob Storage through a gRPC `ClientInterceptor`, with automatic orchestrator response chunking for oversized gRPC messages.

**Overall risk**: Moderate
**Merge recommendation**: Safe with fixes — two performance issues (F1, F2) should be addressed before merge.

### Architecture

This is a well-designed refactoring that moves payload externalization from application-level (manual per-call-site handling) down to the transport layer (gRPC interceptor). Key strengths:
- **Transparency** — application code (orchestrations, activities) is completely unaware of large payloads
- **Separation of concerns** — externalization logic concentrated in one interceptor
- **Extensibility** — `addInterceptor` API is generic, future interceptors can be added
- **Cross-SDK alignment** — chunking logic matches .NET SDK behavior
- **Test quality** — comprehensive coverage with mock unit tests and Azurite-backed integration tests

### Findings by Severity
- High: 2 (F1, F2)
- Medium: 3 (F3, F4, F5)
- Low: 2 (F6, F7)
- Nit: 2 (N1, N2)

### Additional Notes

**[F6] Low / P3 — `maybeResolve` (download path) has no error handling**: `payloadStore.download()` in `maybeResolve` can throw if a blob is deleted or there's a network failure. Unlike the upload path (which has try-catch in `externalizeActivityResponse`/`externalizeOrchestratorResponse`), the download path has no protection. An exception here propagates through `onMessage`, breaking the entire `getWorkItems` streaming call. Consider adding try-catch around download calls to at least log the error and let the orchestration fail explicitly rather than silently disconnecting the stream.

**[N1] Nit — Proto changes include unrelated additions**: `ActivityRequest.tags`, `RewindOrchestrationAction`, `PurgeInstanceFilter.timeout` appear to be upstream proto sync additions unrelated to large payloads. No functional impact, but worth noting in the PR description.

**[N2] Nit — Sample hardcodes token prefix**: `LargePayloadSample.java` line 123 checks `payload.startsWith("blob:v1:")` as a defensive assertion. Acceptable for a sample, but will need updating if the token format changes.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang, line 1291: **[F1] High / P1 — `utf8ByteLength` allocates a large byte array on every call**

`s.getBytes(StandardCharsets.UTF_8).length` allocates a full copy of the string as a byte array just to measure its length. `maybeExternalize` is called on every `StringValue` field in every gRPC message. For strings near the 900KB threshold, each call allocates ~900KB+ of garbage.

**Impact**: Significant GC pressure under high-throughput scenarios.

**Suggestion**: Compute UTF-8 length by iterating chars (zero allocation):
```java
private static int utf8ByteLength(String s) {
    int count = 0;
    for (int i = 0; i < s.length(); i++) {
        char c = s.charAt(i);
        if (c <= 0x7F) count++;
        else if (c <= 0x7FF) count += 2;
        else if (Character.isHighSurrogate(c)) { count += 4; i++; }
        else count += 3;
    }
    return count;
}
```

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
YunchuWang, line 1000: **[F2] High / P1 — `estimateChunkSerializedSize` clones and builds entire protobuf on every iteration**

The inner loop calls `chunk.clone()...build().getSerializedSize()` for each action added. This is O(N²) — each clone copies all previously-accumulated actions. For orchestrations with many actions (100+ `callActivity` fan-outs), this creates significant latency.

**Suggestion**: Cache the envelope overhead size (computed once), then incrementally track size by adding each action's `getSerializedSize()` + ~5 bytes varint overhead. Only do the expensive clone+build when the estimate is close to the limit.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
YunchuWang: **[F3] Medium / P2 — `createIfNotExists` called on every upload**

In steady state the container already exists, but every `upload()` makes an extra HTTP round-trip. Under high-frequency uploads, this adds unnecessary latency.

**Suggestion**: Use an `AtomicBoolean` to skip after first confirmation:
```java
private final AtomicBoolean containerVerified = new AtomicBoolean(false);
// In upload():
if (!containerVerified.get()) {
    containerClient.createIfNotExists();
    containerVerified.set(true);
}
```

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang: **[F4] Medium / P2 — Non-permanent storage exceptions become RuntimeException, may leave orchestrations stuck**

When `externalizeOrchestratorResponse` catches a non-permanent exception (transient network failure), it re-throws as `RuntimeException`. This propagates through gRPC `sendMessage`. The worker's main loop only retries on `StatusRuntimeException` — other exceptions are not gracefully handled. The orchestration won't receive a completion, potentially leaving it stuck until timeout.

**Suggestion**: Convert transient storage failures into a retriable orchestration failure (`isNonRetriable=false`), or wrap `completeOrchestratorTask` in the worker loop with try-catch that abandons the work item.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang, line 260: **[F5] Medium / P2 — `externalizeOrchestratorAction` uses independent `if` blocks for a protobuf `oneof` field**

`OrchestratorAction.orchestratorActionType` is a protobuf `oneof` — exactly one case is active at a time. But this method checks all 6 cases with independent `if` blocks. Functionally correct (only one matches), but wasteful and inconsistent with `resolveEventPayloads` which uses `switch (e.getEventTypeCase())`.

**Suggestion**: Use `switch (action.getOrchestratorActionTypeCase())` for consistency.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadStorageOptions.java
YunchuWang, line 97: **[F7] Low / P3 — No validation that `maxPayloadBytes >= thresholdBytes`**

It's possible to set `maxPayloadBytes` smaller than `thresholdBytes`. Any payload exceeding the threshold would also exceed the max, causing `maybeExternalize` to throw `PayloadStorageException` instead of externalizing — effectively disabling the feature with a misleading error.

**Suggestion**: Validate `maxPayloadBytes >= thresholdBytes` in the `LargePayloadInterceptor` constructor.

YunchuWang (commented):
## Code Review Summary

**Change intent**: Add large payload externalization support to the standalone Durable Task Java SDK via a new `azure-blob-payloads` module. Payloads exceeding a configurable threshold are transparently uploaded to Azure Blob Storage through a gRPC `ClientInterceptor`, with automatic orchestrator response chunking for oversized gRPC messages.

**Overall risk**: Moderate
**Merge recommendation**: Safe with fixes — two performance issues (F1, F2) should be addressed before merge.

### Architecture

This is a well-designed refactoring that moves payload externalization from application-level (manual per-call-site handling) down to the transport layer (gRPC interceptor). Key strengths:
- **Transparency** — application code (orchestrations, activities) is completely unaware of large payloads
- **Separation of concerns** — externalization logic concentrated in one interceptor
- **Extensibility** — `addInterceptor` API is generic, future interceptors can be added
- **Cross-SDK alignment** — chunking logic matches .NET SDK behavior
- **Test quality** — comprehensive coverage with mock unit tests and Azurite-backed integration tests

### Findings by Severity
- High: 2 (F1, F2)
- Medium: 3 (F3, F4, F5)
- Low: 2 (F6, F7)
- Nit: 2 (N1, N2)

### Additional Notes

**[F6] Low / P3 — `maybeResolve` (download path) has no error handling**: `payloadStore.download()` in `maybeResolve` can throw if a blob is deleted or network failure. Unlike the upload path (which has try-catch), the download path has no protection. An exception propagates through `onMessage`, breaking the entire `getWorkItems` streaming call. Consider adding try-catch to let the orchestration fail explicitly rather than silently disconnecting.

**[N1] Nit — Proto changes include unrelated additions**: `ActivityRequest.tags`, `RewindOrchestrationAction`, `PurgeInstanceFilter.timeout` appear to be upstream proto sync. No functional impact but worth noting in PR description.

**[N2] Nit — Sample hardcodes token prefix**: `LargePayloadSample.java` line 123 checks `payload.startsWith("blob:v1:")`. Acceptable for a sample but will need updating if token format changes.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang, line 1291: **[F1] High / P1 — `utf8ByteLength` allocates a large byte array on every call**

`s.getBytes(StandardCharsets.UTF_8).length` allocates a full copy of the string as a byte array just to measure its length. `maybeExternalize` is called on every `StringValue` field in every gRPC message. For strings near the 900KB threshold, each call allocates ~900KB+ of garbage.

**Impact**: Significant GC pressure under high-throughput scenarios.

**Suggestion**: Compute UTF-8 length by iterating chars (zero allocation):
```java
private static int utf8ByteLength(String s) {
    int count = 0;
    for (int i = 0; i < s.length(); i++) {
        char c = s.charAt(i);
        if (c <= 0x7F) count++;
        else if (c <= 0x7FF) count += 2;
        else if (Character.isHighSurrogate(c)) { count += 4; i++; }
        else count += 3;
    }
    return count;
}
```

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java
YunchuWang, line 1000: **[F2] High / P1 — `estimateChunkSerializedSize` clones and builds entire protobuf on every iteration**

The inner loop calls `chunk.clone()...build().getSerializedSize()` for each action added. This is O(N²) — each clone copies all previously-accumulated actions. For orchestrations with many actions (100+ `callActivity` fan-outs), this creates significant latency.

**Suggestion**: Cache the envelope overhead size (computed once), then incrementally track size by adding each action's `getSerializedSize()` + ~5 bytes varint overhead. Only do the expensive clone+build when the estimate is close to the limit.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
YunchuWang: **[F3] Medium / P2 — `createIfNotExists` called on every upload**

In steady state the container already exists, but every `upload()` makes an extra HTTP round-trip. Under high-frequency uploads, this adds unnecessary latency.

**Suggestion**: Use an `AtomicBoolean` to skip after first confirmation:
```java
private final AtomicBoolean containerVerified = new AtomicBoolean(false);
// In upload():
if (!containerVerified.get()) {
    containerClient.createIfNotExists();
    containerVerified.set(true);
}
```

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang: **[F4] Medium / P2 — Non-permanent storage exceptions become RuntimeException, may leave orchestrations stuck**

When `externalizeOrchestratorResponse` catches a non-permanent exception (transient network failure), it re-throws as `RuntimeException`. This propagates through gRPC `sendMessage`. The worker's main loop only retries on `StatusRuntimeException` — other exceptions are not gracefully handled. The orchestration won't receive a completion, potentially leaving it stuck until timeout.

**Suggestion**: Convert transient storage failures into a retriable orchestration failure (`isNonRetriable=false`), or wrap `completeOrchestratorTask` in the worker loop with try-catch that abandons the work item.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
YunchuWang, line 260: **[F5] Medium / P2 — `externalizeOrchestratorAction` uses independent `if` blocks for a protobuf `oneof` field**

`OrchestratorAction.orchestratorActionType` is a protobuf `oneof` — exactly one case is active at a time. But this method checks all 6 cases with independent `if` blocks. Functionally correct (only one matches), but wasteful and inconsistent with `resolveEventPayloads` which uses `switch (e.getEventTypeCase())`.

**Suggestion**: Use `switch (action.getOrchestratorActionTypeCase())` for consistency.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadStorageOptions.java
YunchuWang, line 97: **[F7] Low / P3 — No validation that `maxPayloadBytes >= thresholdBytes`**

It's possible to set `maxPayloadBytes` smaller than `thresholdBytes`. Any payload exceeding the threshold would also exceed the max, causing `maybeExternalize` to throw `PayloadStorageException` instead of externalizing — effectively disabling the feature with a misleading error.

**Suggestion**: Validate `maxPayloadBytes >= thresholdBytes` in the `LargePayloadInterceptor` constructor.

YunchuWang (commented):
## Code Review Summary

**Change intent**: Add large payload externalization support to the standalone Durable Task Java SDK via a new `azure-blob-payloads` module. Payloads exceeding a configurable threshold are transparently uploaded to Azure Blob Storage through a gRPC `ClientInterceptor`, with automatic orchestrator response chunking for oversized gRPC messages.

**Overall risk**: Moderate
**Merge recommendation**: Safe with fixes — two performance issues (F1, F2) should be addressed before merge.

### Architecture

This is a well-designed refactoring that moves payload externalization from application-level (manual per-call-site handling) down to the transport layer (gRPC interceptor). Key strengths:
- **Transparency** — application code (orchestrations, activities) is completely unaware of large payloads
- **Separation of concerns** — externalization logic concentrated in one interceptor
- **Extensibility** — `addInterceptor` API is generic, future interceptors can be added
- **Cross-SDK alignment** — chunking logic matches .NET SDK behavior
- **Test quality** — comprehensive coverage with mock unit tests and Azurite-backed integration tests

### Findings by Severity
- High: 2 (F1, F2)
- Medium: 3 (F3, F4, F5)
- Low: 2 (F6, F7)
- Nit: 2 (N1, N2)

### Additional Notes

**[F6] Low / P3 — `maybeResolve` (download path) has no error handling**: `payloadStore.download()` in `maybeResolve` can throw if a blob is deleted or network failure. Unlike the upload path (which has try-catch), the download path has no protection. An exception propagates through `onMessage`, breaking the entire `getWorkItems` streaming call. Consider adding try-catch to let the orchestration fail explicitly rather than silently disconnecting.

**[N1] Nit — Proto changes include unrelated additions**: `ActivityRequest.tags`, `RewindOrchestrationAction`, `PurgeInstanceFilter.timeout` appear to be upstream proto sync. No functional impact but worth noting in PR description.

**[N2] Nit — Sample hardcodes token prefix**: `LargePayloadSample.java` line 123 checks `payload.startsWith("blob:v1:")`. Acceptable for a sample but will need updating if token format changes.

YunchuWang (commented):
## Code review — High & Medium findings

Focused review of the large-payload externalization + auto-chunking path. 3 High + 6 Medium issues below. Happy to discuss any of these.

---

### [High] F1 — `CompleteOrchestrationAction.carryoverEvents` and `TaskFailureDetails.stackTrace` are not externalized
**File:** `azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java` — `externalizeOrchestratorAction`, `COMPLETEORCHESTRATION` branch.

`CompleteOrchestrationAction` has two payload surfaces beyond `result`/`details` that can legitimately grow very large:
1. `repeated HistoryEvent carryoverEvents` (continue-as-new buffers `EventRaised` / `ExecutionStarted` payloads for the next instance).
2. `TaskFailureDetails.stackTrace` / `errorMessage` (effectively unbounded user content).

Only `result` and `details` go through `maybeExternalize`. When `supportsLargePayloads=true`, `completeOrchestratorTaskWithChunking` also **skips** the pre-send size validator, so an oversized `CompleteOrchestrationAction` (large carryover events on continue-as-new) will hit the wire and fail with `RESOURCE_EXHAUSTED`. The orchestration then loops and re-fails on every replay.

**Recommendation:** Walk `carryoverEvents` through the same event-payload externalization used on the inbound path, and run `failureDetails.getStackTrace()` through `maybeExternalize`. Apply the same treatment to `TaskFailureDetails` on `ActivityResponse` failures.

---

### [High] F2 — Transient blob-resolve failure during WorkItem ingestion creates an unrecoverable replay loop
**File:** `LargePayloadInterceptor.java` — `onMessage` wrapper (response path).

Any exception in `resolveResponsePayloads` is converted to `Status.INTERNAL` and thrown from `onMessage`. For the streaming `GetWorkItems` RPC, this terminates the stream; the worker reconnects; the sidecar re-dispatches the same work item; we re-download the same (still-failing) blob; we crash again. No circuit breaker, no backoff, no way to fail a single poisoned WorkItem out.

This applies both to:
- **Transient** failures (429 / 5xx after 8-retry exhaustion), and
- **Permanent** failures (404 due to retention/cleanup/container rename, malformed token, container mismatch).

For permanent 404s the only recovery is a manual purge — a real foot-gun in production.

**Recommendation:**
- For permanent failures (404, malformed token, container mismatch) complete the work item as non-retriable failed via `completeActivityTask` / `completeOrchestratorTask` with `TaskFailureDetails.isNonRetriable=true` rather than throwing from the stream listener.
- For transient failures use `Status.UNAVAILABLE` (not `INTERNAL`) so sidecar-side backoff applies.

---

### [High] F3 — `isKnownPayloadToken` matches any user string starting with `blob:v1:`, causing false-positive download attempts
**File:** `azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java` — `isKnownPayloadToken`.

The only check is `value.startsWith(TOKEN_PREFIX)`. Any user-provided JSON value whose content happens to start with `blob:v1:` (configuration data, a URL fragment, a test fixture, or an attacker-controlled `RaiseEvent` input) will be treated as a token on the response path. `download` then either:
1. Hits the container-mismatch check and throws `IllegalArgumentException` → flows into F2 and poisons the work item forever, or
2. If the user string happens to match the configured container, issues a blob GET for an attacker-chosen blob name. Disclosure is impractical (UUID-128 blob names) but DoS via a crafted event input is.

**Recommendation:** Validate the full token grammar — require `blob:v1:<container>:<32-hex-chars>` — not just the prefix. Also reduces the blast radius of F2.

---

### [Medium] F4 — Pre-send action-size validation skipped entirely when `supportsLargePayloads=true`, even for non-externalized fields
**File:** `client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcWorker.java` — `completeOrchestratorTaskWithChunking`.

The guard is `if (!this.supportsLargePayloads)` — i.e. trust the interceptor to shrink everything. But the interceptor only externalizes specific `StringValue` fields (result, input, details, reason, customStatus, …). A `CompleteOrchestrationAction` with huge `carryoverEvents` (see F1) or large `failureDetails` will not be reduced by the interceptor and will not be caught by the validator either.

**Recommendation:** Always run `validateActionsSize`. The interceptor sits on `sendMessage` and will already have had its chance by the time the blocking stub invokes gRPC. Alternatively, only bypass validation once every payload-carrying field is confirmed externalizable.

---

### [Medium] F5 — New Java chunking depends exclusively on proto-deprecated `isPartial` / `chunkIndex`
**File:** `internal/durabletask-protobuf/protos/orchestrator_service.proto` + `DurableTaskGrpcWorker.completeOrchestratorTaskWithChunking`.

The proto comment in this PR says `/* Chunking logic has since been deprecated and fields related to it are marked as such */`, and both fields are `[deprecated=true]`. The new Java code depends entirely on them. If a newer sidecar drops or ignores these fields, chunked responses will be silently treated as a completion on chunk 0 and subsequent chunks dropped — very hard to diagnose.

**Recommendation:** Confirm with DTS backend owners that:
1. Deprecated-but-supported semantics will hold for the supported lifecycle, and
2. The long-term replacement is large-payload externalization (shipped in this same PR), with chunking as a fallback for users who can't configure Blob.

Add a javadoc comment in the Java code explicitly calling out the deprecation and the intended horizon.

---

### [Medium] F6 — `BlobPayloadStore.upload` uses unconditional overwrite
**File:** `BlobPayloadStore.java` — both compressed (`uploadWithResponse` with `null` requestConditions) and uncompressed (`blob.upload(stream, len, true)`) paths.

Blob names are random UUIDs so collision is astronomically unlikely, but an `If-None-Match: *` precondition adds defense-in-depth at zero operational cost — e.g. against a future bug where a caller-supplied `PayloadStore` generates deterministic names, or a future refactor.

**Recommendation:** Use `BlobRequestConditions.setIfNoneMatch("*")` on uploads and convert 409 to a hard failure.

---

### [Medium] F7 — Inconsistent null handling across `LargePayloadStorageOptions` setters
**File:** `LargePayloadStorageOptions.java`.

`setConnectionString(null)` silently converts to `""`, but `setCredential(null)` and `setAccountUri(null)` store `null`. A caller cannot reliably clear a previously-set connection string to switch to identity auth.

**Recommendation:** Pick one behavior and apply it to all three. Simplest is to accept `null` on all three and normalize in `BlobPayloadStore` where `hasConnectionString` / `hasIdentityAuth` is computed.

---

### [Medium] F8 — `containerVerified` uses `get` then `set` instead of `compareAndSet`
**File:** `BlobPayloadStore.java` — lazy container-ensure path.

Multiple concurrent first-uploads can each call `createIfNotExists()`. It's idempotent so this is harmless today, but the pattern is a classic TOCTOU that static analyzers (SpotBugs) may flag.

**Recommendation:**
```java
if (this.containerVerified.compareAndSet(false, true)) {
    try { this.containerClient.createIfNotExists(); }
    catch (BlobStorageException e) {
        if (e.getStatusCode() != 409) {
            this.containerVerified.set(false); // allow retry
            throw new PayloadStorageException(...);
        }
    }
}
```

---

### [Medium] F9 — `maxChunkSizeBytes` default of 3.9 MB has no safety margin over gRPC's default 4 MB
**File:** `DurableTaskGrpcWorkerBuilder.java` — `maxChunkSizeBytes = 4_089_446`.

The size check uses `response.getSerializedSize()` (protobuf payload only). gRPC framing adds 5 bytes (compression flag + length prefix) and HTTP/2 framing adds more. The chunking loop also uses a 0.99 multiplier before falling to the accurate check, so chunks can approach the full 3.9 MB. If the sidecar runs with the gRPC default `maxInboundMessageSize` (4 MB = 4,194,304 B) the margin is <100 KB.

**Recommendation:** Either drop the default to ~3.5 MB, or document clearly in javadoc that the sidecar must accept ≥4 MB inbound messages and link to the .NET constant so the two SDKs can't drift silently.

---

### Out of scope for this comment set (tracked locally, not posted)
- F10 client-side `sendMessage` not wrapping upload errors.
- F11 `azuremanaged` test dependency.
- F12 unrelated proto additions (`ActivityRequest.tags`, `RewindOrchestrationAction`, `PurgeInstanceFilter.timeout`).
- F13 stray `;;` on `chunkIndex`.
- F14 `estimateChunkSerializedSize` clone-per-check performance.

Happy to elaborate on any of these or pair on the fixes.

samples/src/main/java/io/durabletask/samples/LargePayloadSample.java
```diff
@@ -0,0 +1,200 @@
...
+ * <ul>
+ *   <li>{@code DURABLE_TASK_SCHEDULER_CONNECTION_STRING} — DTS connection string
+ *       (default: {@code Endpoint=http://localhost:8080;TaskHub=default;Authentication=None})</li>
+ *   <li>{@code PAYLOAD_STORAGE_CONNECTION_STRING} — Azure Storage connection string
+ *       (default: {@code UseDevelopmentStorage=true})</li>
+ *   <li>{@code PAYLOAD_SIZE_BYTES} — payload size in bytes (default: 1572864 = 1.5 MiB)</li>
+ *   <li>{@code EXTERNALIZE_THRESHOLD_BYTES} — externalization threshold (default: 900000)</li>
+ * </ul>
+ */
+final class LargePayloadSample {
```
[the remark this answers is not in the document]
bachuv: It's runnable from gradlew :samples:runLargePayloadSample.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/BlobPayloadStore.java
```diff
@@ -0,0 +1,321 @@
...
+     * the check is skipped on all future calls. On failure the guard is reset so a later
+     * call can retry.
+     */
+    private void ensureContainerExists() {
+        if (this.containerVerified) {
+            return;
+        }
+
+        CountDownLatch latch = new CountDownLatch(1);
+        CountDownLatch existing = this.containerLatch.compareAndExchange(null, latch);
```
YunchuWang: **High:** This uses AtomicReference.compareAndExchange, which was added in Java 9. The new module targets Java 8 (sourceCompatibility = JavaVersion.VERSION_1_8), and compiling on JDK 11 with 	argetCompatibility still emits a call to the Java 9+ method, so Java 8 consumers can hit NoSuchMethodError. Please replace this with Java 8-compatible compareAndSet logic.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
```diff
@@ -0,0 +1,1233 @@
...
+        }
+        String resolved = this.payloadStore.download(strValue);
+        return StringValue.of(resolved);
+    }
+
+    /**
+     * Determines whether an exception represents a permanent storage failure.
+     */
+    static boolean isPermanentStorageFailure(Exception ex) {
+        if (ex instanceof PayloadStorageException) {
```
YunchuWang, line 1156: **High:** Classifying every PayloadStorageException as permanent makes transient Azure Storage failures non-retriable. BlobPayloadStore wraps all BlobStorageExceptions, including 408/429/5xx/network-timeout cases, as PayloadStorageException, so a temporary storage outage can mark activities/orchestrations as non-retriable or poison-complete work items. Please preserve the underlying status/cause and classify only true permanent cases as non-retriable.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
```diff
@@ -0,0 +1,1233 @@
...
+                break;
+            }
+            default:
+                break;
+        }
+
+        return builder != null ? builder.build() : action;
+    }
+
+    private Object externalizeEntityBatchResult(EntityBatchResult r) {
```
YunchuWang, line 365: **Medium:** xternalizeEntityBatchResult is not wrapped in the same permanent/transient failure handling used for activity and orchestrator responses. If externalizing entity state/results/actions fails permanently, the exception bubbles out through completeEntityTask, the worker catch path abandons the entity work item, and the sidecar can keep redispatching it instead of recording a non-retriable entity failure. Please convert permanent entity externalization failures into an EntityBatchResult failure completion.

azure-blob-payloads/src/main/java/com/microsoft/durabletask/azureblobpayloads/LargePayloadInterceptor.java
```diff
@@ -0,0 +1,1233 @@
...
+
+        StringValue entityState = maybeExternalize(r.getEntityState());
+        if (entityState != r.getEntityState()) {
+            builder.setEntityState(entityState);
+            changed = true;
+        }
+
+        for (int i = 0; i < r.getResultsCount(); i++) {
+            OperationResult result = r.getResults(i);
+            if (result.hasSuccess()) {
```
YunchuWang: **Medium:** Entity failure payloads are not externalized here. This handles entity state, successful operation results, and outgoing entity actions, but skips EntityBatchResult.failureDetails and per-operation OperationResultFailure.failureDetails; large entity exception stack traces can still be sent inline and exceed gRPC limits. Please apply the same xternalizeFailureDetails recursion used for activity/orchestration failures.

azure-blob-payloads/build.gradle
```diff
@@ -0,0 +1,116 @@
...
+def exeSuffix = isWindows ? ".exe" : ""
+
+dependencies {
+    api project(':client')
+
+    // Azure Storage Blobs
+    implementation "com.azure:azure-storage-blob:${azureStorageBlobVersion}"
+
+    // TokenCredential abstraction (from azure-core)
+    implementation "com.azure:azure-core:${azureCoreVersion}"
```
YunchuWang: **Medium:** LargePayloadStorageOptions exposes TokenCredential in the public API, but zure-core is declared as implementation, so Gradle consumers of this library will not get TokenCredential on their compile classpath unless they add azure-core manually. This should be an pi dependency. Also, since this is a new consumable module, please add the same Maven publication setup/version alignment as the other SDK artifacts so it is published consistently.

YunchuWang (commented):
High/medium code review findings for large payload support.

YunchuWang (approved):
Approving — H1 (container-creation race) and M1 (threshold propagation) are fixed in 38ce81a, and the latest commits include several nice unsolicited improvements:

- `isPermanentStorageFailure` now unwraps `PayloadStorageException` to inspect the underlying `BlobStorageException` status, so transient blob failures are correctly classified as transient.
- `externalizeEntityBatchResult` now has the same try/catch + permanent/transient mapping as the activity/orchestrator paths, closing a parity gap.
- `stringValueSavings` now uses UTF-8 byte length (matching the interceptor), so the worker's pre-send estimate and the interceptor's externalization decision use consistent units.

Nit (non-blocking): the `## Unreleased` section in CHANGELOG.md is still empty — please add a line for this PR before/at merge.

Follow-ups for a future PR (none blocking):
- `download()` doesn't enforce `maxPayloadBytes` on the inbound side.
- `tryFailWorkItem` performs a blocking unary RPC inside `onMessage`; consider hopping to an executor to keep the gRPC inbound thread responsive under load.
- `setContainerName` accepts any string; a light Azure-naming-rules check would catch typos at config time instead of first upload.
- The chunking envelope-size check could be done upfront against `maxChunkSizeBytes` to fail fast.
- `encodeToken` could validate inputs (container/blob name shape) defensively.

Nice work overall — the H1 fix in particular is a clean `CountDownLatch` + `AtomicReference` pattern.

PR #292, opened 2026-07-01
Add client APIs to list instance IDs and read orchestration history

bachuv:

Adds client APIs to enumerate orchestration instances and read a completed instance's
history, plus a public history-event model.

#### Changes
- **`DurableTaskClient.listInstanceIds(ListInstanceIdsQuery)`** → `ListInstanceIdsResult` — page terminal/instance IDs by filter (completion window, statuses, continuation token).
- **`DurableTaskClient.getOrchestrationHistory(String instanceId)`** → `List<HistoryEvent>` — retrieve an instance's full ordered history.
- **New public model** `com.microsoft.durabletask.history` — `HistoryEvent` plus per-type subclasses (`ExecutionStarted/Completed`, `TaskScheduled/Completed/Failed`, `Timer*`, `SubOrchestration*`, `EntityOperation*`/`EntityLock*`, `EventSent/Raised`, `Generic`, `HistoryState`, …).
- Backed by the `ListInstanceIds` and `StreamInstanceHistory` gRPC operations (managed DTS; sidecar/emulator ≥ v0.4.22).

##### Notes
- Foundation for a follow-up durable history-export feature (stacked PR).

### Pull request checklist

### DIFF
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java
@@ -248,13 +248,15 @@ public abstract class DurableTaskClient implements AutoCloseable {
     }
 
     /**
-     * Gets the full history of an orchestration instance as an ordered list of {@link HistoryEvent} objects.
+     * Retrieves the history of the specified orchestration instance as a list of {@link HistoryEvent} objects.
      * <p>
      * The events are returned in execution order. This is useful for archiving or offline analysis of an instance's
      * execution history. Use {@code instanceof} to inspect each concrete event type.
      *
-     * @param instanceId the unique ID of the orchestration instance whose history to fetch
-     * @return the instance's history events in order; empty if the instance has no history
+     * @param instanceId the instance ID of the orchestration
+     * @return the list of {@link HistoryEvent} objects representing the orchestration's history, in execution order;
+     *         empty if the instance has no history
+     * @throws IllegalArgumentException if {@code instanceId} is null
      * @throws UnsupportedOperationException if the current client implementation does not support history retrieval
      */
     public List<HistoryEvent> getOrchestrationHistory(String instanceId) {
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java
@@ -240,8 +240,12 @@ public abstract class DurableTaskClient implements AutoCloseable {
      *
      * @param query filter criteria: completion-time window, terminal runtime statuses, page size, and pagination cursor
      * @return a page of matching instance IDs and a cursor for the next page
+     * @throws UnsupportedOperationException if the current client implementation does not support listing instance IDs
      */
-    public abstract ListInstanceIdsResult listInstanceIds(ListInstanceIdsQuery query);
+    public ListInstanceIdsResult listInstanceIds(ListInstanceIdsQuery query) {
+        throw new UnsupportedOperationException(
+                "Listing instance IDs is not supported by this client implementation.");
+    }
 
     /**
      * Gets the full history of an orchestration instance as an ordered list of {@link HistoryEvent} objects.
@@ -251,8 +255,12 @@ public abstract class DurableTaskClient implements AutoCloseable {
      *
      * @param instanceId the unique ID of the orchestration instance whose history to fetch
      * @return the instance's history events in order; empty if the instance has no history
+     * @throws UnsupportedOperationException if the current client implementation does not support history retrieval
      */
-    public abstract List<HistoryEvent> getOrchestrationHistory(String instanceId);
+    public List<HistoryEvent> getOrchestrationHistory(String instanceId) {
+        throw new UnsupportedOperationException(
+                "Retrieving orchestration history is not supported by this client implementation.");
+    }
 
     /**
      * Initializes the target task hub data store.
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java
@@ -2,8 +2,11 @@
 // Licensed under the MIT License.
 package com.microsoft.durabletask;
 
+import com.microsoft.durabletask.history.HistoryEvent;
+
 import javax.annotation.Nullable;
 import java.time.Duration;
+import java.util.List;
 import java.util.concurrent.TimeoutException;
 
 /**
@@ -225,6 +228,31 @@ public abstract class DurableTaskClient implements AutoCloseable {
      * @return the result of the query operation, including instance metadata and possibly a continuation token
      */
     public abstract OrchestrationStatusQueryResult queryInstances(OrchestrationStatusQuery query);
+
+    /**
+     * Lists the IDs of terminal orchestration instances that completed within a time window.
+     * <p>
+     * Unlike {@link #queryInstances(OrchestrationStatusQuery)}, which filters by creation time and returns full
+     * metadata, this method filters by <em>completion</em> time and returns only instance IDs, making it efficient
+     * for bulk enumeration such as archival/export. Results are paged; pass
+     * {@link ListInstanceIdsResult#getContinuationToken()} back via
+     * {@link ListInstanceIdsQuery#setContinuationToken(String)} to fetch subsequent pages.
+     *
+     * @param query filter criteria: completion-time window, terminal runtime statuses, page size, and pagination cursor
+     * @return a page of matching instance IDs and a cursor for the next page
+     */
+    public abstract ListInstanceIdsResult listInstanceIds(ListInstanceIdsQuery query);
+
+    /**
+     * Gets the full history of an orchestration instance as an ordered list of {@link HistoryEvent} objects.
+     * <p>
+     * The events are returned in execution order. This is useful for archiving or offline analysis of an instance's
+     * execution history. Use {@code instanceof} to inspect each concrete event type.
+     *
+     * @param instanceId the unique ID of the orchestration instance whose history to fetch
+     * @return the instance's history events in order; empty if the instance has no history
+     */
+    public abstract List<HistoryEvent> getOrchestrationHistory(String instanceId);
 
     /**
      * Initializes the target task hub data store.
--- a/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
+++ b/client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
@@ -299,6 +299,7 @@ public final class DurableTaskGrpcClient extends DurableTaskClient {
 
     @Override
     public ListInstanceIdsResult listInstanceIds(ListInstanceIdsQuery query) {
+        Helpers.throwIfArgumentNull(query, "query");
         ListInstanceIdsRequest.Builder builder = ListInstanceIdsRequest.newBuilder();
         Optional.ofNullable(query.getCompletedTimeFrom()).ifPresent(from -> builder.setCompletedTimeFrom(DataConverter.getTimestampFromInstant(from)));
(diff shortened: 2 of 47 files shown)


client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java
```diff
@@ -226,6 +229,31 @@ public abstract OrchestrationMetadata waitForInstanceCompletion(
...
+     * Unlike {@link #queryInstances(OrchestrationStatusQuery)}, which filters by creation time and returns full
+     * metadata, this method filters by <em>completion</em> time and returns only instance IDs, making it efficient
+     * for bulk enumeration such as archival/export. Results are paged; pass
+     * {@link ListInstanceIdsResult#getContinuationToken()} back via
+     * {@link ListInstanceIdsQuery#setContinuationToken(String)} to fetch subsequent pages.
+     *
+     * @param query filter criteria: completion-time window, terminal runtime statuses, page size, and pagination cursor
+     * @return a page of matching instance IDs and a cursor for the next page
+     */
+    public abstract ListInstanceIdsResult listInstanceIds(ListInstanceIdsQuery query);
```
nytian: Just wondering  which classes extend DurableTaskClient? Only within this repo, right? Can we add a default implementation that throws UnsupportedOperationException? I think its safer for these breaking changes.

client/src/main/java/com/microsoft/durabletask/ListInstanceIdsQuery.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Sets the maximum number of instance IDs to return per page. The default value is 100.
+     * <p>
+     * A page may contain fewer IDs than the page size even when more results exist; always use
+     * {@link ListInstanceIdsResult#getContinuationToken()} to determine whether to continue paging.
+     *
+     * @param pageSize the maximum number of instance IDs to return per page
+     * @return this query object
+     */
+    public ListInstanceIdsQuery setPageSize(int pageSize) {
```
nytian, line 76: Lets add a check here to ensure pageSize won't be negative or 0

client/src/test/java/com/microsoft/durabletask/HistoryEventConverterTest.java
```diff
@@ -0,0 +1,226 @@
...
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * Unit tests for {@link HistoryEventConverter}, which maps protobuf history events to the public
+ * {@code com.microsoft.durabletask.history} domain model.
+ */
+public class HistoryEventConverterTest {
```
nytian, line 55: Seems there are some history events are not tested here. Can you add them? Eg the entity types and suspend/resume and continueasnew etc,

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
```diff
@@ -295,6 +297,37 @@ private OrchestrationStatusQueryResult toQueryResult(QueryInstancesResponse quer
...
+                .setInstanceId(instanceId)
+                .build();
+        List<HistoryEvent> historyEvents = new ArrayList<>();
+        Iterator<HistoryChunk> chunks = this.sidecarClient.streamInstanceHistory(request);
+        while (chunks.hasNext()) {
+            for (OrchestratorService.HistoryEvent protoEvent : chunks.next().getEventsList()) {
+                historyEvents.add(HistoryEventConverter.fromProto(protoEvent));
+            }
+        }
+        return historyEvents;
```
nytian, line 345: nit: if cx has a very large history list, we have to return to them after processing them all in memory. So I fell like Stream/Page is better here, but I feel this is not super common case, so up to your decision :)

nytian (commented):
Left a few comments. Thanks for the PR!

client/src/main/java/com/microsoft/durabletask/DurableTaskGrpcClient.java
bachuv: Keeping as is to stay at parity with the existing .NET (IList<HistoryEvent>) and Python (list[HistoryEvent]) APIs

client/src/test/java/com/microsoft/durabletask/HistoryEventConverterTest.java
bachuv: Added

client/src/main/java/com/microsoft/durabletask/DurableTaskClient.java
bachuv: Fixed - removed abstract and the methods now throw UnsupportedOperationException as a default.

YunchuWang (commented):
## Summary

Strong, well-structured PR. The domain model is clean and immutable, and the protobuf→model converter is exactly correct — I verified all 28 `HistoryEvent` oneof cases (fields 3–30) and sampled event fields against `orchestrator_service.proto`, and every mapping and null-guard lines up. Test coverage is good. A handful of things worth resolving before merge, none catastrophic.

## Strengths

- **Converter correctness**: all 28 event types handled; `stringOrNull(hasX, getX)` correctly guards the `StringValue`-wrapped optional fields. One unit test per event type (608 lines), including null-handling and the `EVENTTYPE_NOT_SET → IllegalArgumentException` branch.
- **Model quality**: event classes are `final`, fully immutable, `@Nullable`-annotated, well-documented; `tags` is defensively copied (`Collections.unmodifiableMap(new HashMap<>(tags))`, null→`emptyMap`).
- **Back-compat**: base-class defaults throw `UnsupportedOperationException`, so external `DurableTaskClient` subclasses still compile; the sole impl (`DurableTaskGrpcClient`) overrides both.
- `ListInstanceIdsQuery` validates `pageSize ≥ 1` and defensively copies the status list; `ListInstanceIdsResult` returns an unmodifiable list.

## Findings (by severity)

**1. `getOrchestrationHistory` — forward-compat + stream cleanup (Medium).** The converter throws on any unrecognized event type. A future server adding a new event type makes the *entire* history read fail with `UnsupportedOperationException`. Worse, the blocking gRPC server-streaming iterator is drained in a `while(hasNext)` loop with no `try/finally`; if `fromProto` throws mid-stream the loop exits by exception and the streaming RPC is abandoned without cancellation — a known grpc-java leak pattern. Recommend (a) deciding forward-compat behavior (skip / `GenericEvent` fallback vs. fail-fast), and (b) cancelling the call on early exit.

**2. Whole history materialized in memory (Medium / design).** The API returns `List<HistoryEvent>`, buffering the full history even though the server streams it in chunks — this defeats streaming for very large histories. For the archival/export use case, consider an `Iterable`/streaming/callback shape, or explicitly document the in-memory materialization.

**3. Pagination termination — null vs. empty token (Low–Medium).** `continuationToken` becomes `null` only when the server *omits* `lastInstanceKey`. If the server instead returns a present-but-empty `StringValue` on the last page, the client hands back `""` (non-null); a caller looping on `!= null` re-requests with `""`, which restarts from the beginning → potential infinite paging loop. Recommend treating empty as end-of-results: `hasLastInstanceKey() && !value.isEmpty() ? value : null`. No multi-page pagination test exists to catch this.

**4. Public constructors on ~30 model types (Low / API design).** Because `HistoryEventConverter` sits in `com.microsoft.durabletask` while the model lives in `...history`, every event constructor must be `public` — pushing long positional constructors (e.g. `ExecutionStartedEvent` has 11 params) into the permanent public API even though users only ever *read* these. Moving the converter into the `history` package (or a package-private factory there) would let the constructors be package-private, keeping the public surface read-only.

**5. Naming (Low).** `listInstanceIds` doesn't convey "terminal instances filtered by completion time." The Javadoc clarifies, but the method name is the real contract and it's still unreleased — consider something like `listCompletedInstanceIds`.

**6. CHANGELOG not updated (Low).** New public API but no `Unreleased` entry; the PR checklist item is unchecked.

**7. Unrelated scope (Nit).** `<h3>`→`<h2>` Javadoc edits in `EntityQueryPageable`, `TypedEntityMetadata`, `TypedEntityQueryPageable` (doclint heading-order fixes) are harmless but unrelated to this feature.

## Questions for the author

- Intended behavior when a client hits an event type it doesn't recognize (older client / newer server) — fail the whole read, or skip?
- Is `List` the intended long-term shape, or will a follow-up expose the underlying stream for large histories?

bachuv:
Thank you for the feedback @YunchuWang.

## Changed

**#1b - stream cleanup (the leak).** Fixed. `getOrchestrationHistory` now binds the server-streaming drain to a `Context.CancellableContext` and cancels it in a `finally`, so the RPC is released even if conversion throws mid-stream. This matches .NET's `using AsyncServerStreamingCall<…>` in `GetOrchestrationHistoryAsync`.

**#3 - pagination null vs. empty token.** Fixed. `listInstanceIds` now treats an absent **or** present-but-empty `lastInstanceKey` as end-of-results (`hasLastInstanceKey() && !value.isEmpty()`), so a caller looping on `!= null` can't restart from the beginning. Added 3 in-process-server unit tests (empty / absent / non-empty token). Note: Python has the same latent gap (it only checks `HasField`), so this actually makes Java the most robust of the three while still honoring the shared "null token = done" contract.

**#6 - CHANGELOG.** Added an `Unreleased` entry for the new client APIs (#292).

**Javadoc alignment.** Reworded the `getOrchestrationHistory` doc to match .NET's `GetOrchestrationHistoryAsync` (summary/`@param`/`@return`, plus `@throws IllegalArgumentException`≙`ArgumentNullException` and `@throws UnsupportedOperationException`≙`NotSupportedException`).

## Left as-is - parity with .NET and Python

**#1a - forward-compat on unknown event types.** Kept fail-fast. All three SDKs fail-fast today: .NET throws `NotSupportedException`, Python raises `ValueError`, Java throws `UnsupportedOperationException`. Adding a skip/`UnknownHistoryEvent` fallback would make Java the only SDK that silently tolerates unknown events.

**#2 - `List` vs. streaming.** Kept `List<HistoryEvent>`. .NET returns `IList<HistoryEvent>` and Python returns `List[HistoryEvent]`, both fully materialized. A streaming/`Iterable` overload is reasonable but should land in all three SDKs together, so I've left the shape as-is for now.

**#4 - public constructors.** Kept public. Both peers expose publicly-constructible history events: .NET reuses the public `DurableTask.Core` history types, Python uses public `@dataclass` types so Java's public constructors are at parity. Also, the suggested "move the converter into `history` + package-private ctors" has a hidden cost: `HistoryEventConverter` depends on package-private members of `com.microsoft.durabletask` (`DataConverter.getInstantFromTimestamp`, the `FailureDetails(TaskFailureDetails)` ctor, `OrchestrationRuntimeStatus.fromProtobuf`), so moving it would force those internals to become `public` — trading ~30 read-model ctors for leaking core helpers. Net-negative, so I left it.

**#5 - naming.** Kept `listInstanceIds`. It already matches .NET's `ListInstanceIdsAsync` and Python's `list_instance_ids`; `listCompletedInstanceIds` would break that naming parity for an otherwise-identical method. The Javadoc spells out the "terminal instances by completion time" contract.

## #7 - unrelated Javadoc edits

Keeping these. Reverting the three `<h3>`→`<h2>` files to `main` makes `./gradlew :client:javadoc` **fail** with 3 "heading used out of sequence" errors (javadoc runs at publish time via `withJavadocJar()`). They fix a pre-existing latent doclint break that this PR surfaces because it adds a lot of new public Javadoc.

## Answers to your questions

**Q: Unknown event type — fail the whole read, or skip?** For now, fail-fast, matching .NET and Python.

**Q: Is `List` the long-term shape?** `List` is the intended shape for this PR (parity with .NET/Python).

YunchuWang (dismissed):

YunchuWang (approved):

PR #294, opened 2026-07-17
Enable Custom Exception Provider with Durable Functions Java

nytian:
Today when a Java activity fails, only the exception type and message reach the Durable extension. The TaskFailureDetails from the worker cannot carry user context like error codes or correlation IDs, so that data is dropped at the worker to host boundary. Users have no supported way to surface it in their error handling.

This PR adds an `ActivityMiddleware` that fixes this. When an activity throws, it runs a registered ExceptionPropertiesProvider and serializes the exception into structured TaskFailureDetails JSON, including error type, message, stack trace, nested causes, and custom properties. It runs worker side while the live Exception still exists, which is the only place this data is available before serialization.

Nested causes are emitted as innerFailure, bounded by MAX_INNER_FAILURE_DEPTH (10). Provider discovery uses cross class loader SPI lookup so it works from worker threads. When no provider is registered, the original exception is rethrown unchanged, so the change is additive. Also registers the middleware via SPI and adds the `properties` field to TaskFailureDetails. Adds ActivityMiddlewareTest, and the e2e test at durable extension repo accordingy


azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,373 @@
...
+
+/**
+ * Durable Function Activity Middleware.
+ *
+ * <p>When an activity function throws, this middleware gives a registered
+ * {@link ExceptionPropertiesProvider} the chance to attach custom properties to the failure. If the
+ * provider returns any properties, the exception is reshaped into a serialized
+ * {@code TaskFailureDetails} JSON payload (matching the protobuf JSON shape) so the Durable Task
+ * host extension can surface the structured properties on {@code FailureDetails.Properties}. This
+ * mirrors the {@code durable-functions} JavaScript SDK's activity handler wrapper.
```
bachuv: Can we remove the reference to the durable-functions JavaScript SDK?

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,373 @@
...
+ *
+ * <p>If no provider is registered, or it yields no properties for the thrown exception, the original
+ * exception is re-thrown untouched so the legacy failure behavior is preserved.
```
bachuv:
```suggestion
 * exception is re-thrown untouched.
```

nit: remove the mention of legacy behavior

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,373 @@
...
+ *
+ * <p>The provider is discovered via {@link ServiceLoader} (SPI): an application registers its
+ * implementation in {@code META-INF/services/com.microsoft.durabletask.ExceptionPropertiesProvider}.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
+ * at any time.
```
bachuv: nit: remove these comments
```suggestion
```

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,373 @@
...
+ */
+public class ActivityMiddleware implements Middleware {
+
+    private static final String ACTIVITY_TRIGGER = "DurableActivityTrigger";
+    private static final int MAX_INNER_FAILURE_DEPTH = 10;
```
bachuv: Does this match the depth configured in the other SDKs?

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,373 @@
...
+            chain.doNext(context);
+        } catch (Exception e) {
+            ExceptionPropertiesProvider provider = getProvider();
+            if (provider == null) {
+                throw e;
+            }
+
+            Throwable userException = unwrap(e);
+            Map<String, Object> properties = safeGetProperties(provider, userException);
+            if (properties == null || properties.isEmpty()) {
```
bachuv: Should we also check inner exceptions for custom properties when the outer exception has none?

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
nytian: I remember .NET has a similar configuration,but the max depth value may be different. And for dotnet, it applies more broadly to all TaskFailedException scenarios, not just activity functions. Since this change only affects activity functions, I think a limit of 10 should be sufficient. It just helps ensure we don't end up recursing indefinitely.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
nytian: yeah  Imissed that. We do support inner failures for at dotnet and js, let me update this part..

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
bachuv: > I remember .NET has a similar configuration,but the max depth value may be different.

@nytian, what is the max depth value for .NET?

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
bachuv, line 1: In this file, instead of writing the JSON by hand, can we reuse the existing FailureDetails.fromException(...) and serialize its toProto() with protobuf's JsonFormat? That drops ~180 lines and keeps the output in sync with the real TaskFailureDetails schema automatically.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,369 @@
...
+        return hasCustomProperties;
+    }
+
+    /**
+     * Serializes a single property value as JSON, handling strings, booleans, numbers (non-finite
+     * doubles fall back to strings), maps, iterables, and arrays; anything else is written as its
+     * {@code toString()}.
+     */
+    @SuppressWarnings("unchecked")
+    private static void appendValue(StringBuilder sb, Object value) {
```
bachuv: appendValue handles Object[] but not primitive arrays. In Java, an int[] is not an Object[], so it skips every branch and falls through to toString() producing garbage like "[I@1a2b3c" in the JSON.

Can we add a branch that detects primitive arrays with value.getClass().isArray() and iterates them using java.lang.reflect.Array.getLength(value) / Array.get(value, i), so each element gets serialized properly. Since a provider can return an int[]/long[]/etc. in its properties map, this closes the gap. Same fix would apply to FailureDetails.convertToProtoValue if we ever share that code.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,369 @@
...
+        boolean hasCustomProperties = properties != null && !properties.isEmpty();
+
+        sb.append('{');
+        sb.append("\"errorType\":");
+        appendString(sb, exception.getClass().getName());
+        sb.append(",\"errorMessage\":");
+        appendString(sb, exception.getMessage() != null ? exception.getMessage() : "");
+        sb.append(",\"stackTrace\":");
+        appendString(sb, getFullStackTrace(exception));
+        sb.append(",\"isNonRetriable\":false");
```
bachuv: How do we know that this should be hard coded to false and isn't true in any scenarios?

azurefunctions/src/test/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddlewareTest.java
```diff
@@ -0,0 +1,367 @@
...
+        });
+
+        BusinessException original = new BusinessException("boom");
+        ActivityMiddleware middleware = new ActivityMiddleware();
+
+        Exception thrown = assertThrows(Exception.class,
+                () -> middleware.invoke(activityContext(), throwingChain(original)));
+
+        // The original exception is replaced by a structured-failure carrier whose message is JSON.
+        assertNotSameInstance(original, thrown);
```
bachuv: Can we use the built-in `assertNotSame(original, thrown)` here?

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
nytian: We don't have that for .net, its more likely bounded by grpc message size

azurefunctions/src/test/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddlewareTest.java
nytian: udpated

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
nytian: Good point. Hard-code is not a good thing. Updated

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -204,7 +208,13 @@ static String getFullStackTrace(Throwable e) {
...
     }
 
-    TaskFailureDetails toProto() {
+    /**
+     * Converts this failure to its protocol representation.
+     *
+     * @return the protocol representation of this failure
+     */
+    @Nonnull
+    public TaskFailureDetails toProto() {
```
bachuv, line 217: Why is this public? If it's used by tests, then can we make it internal?

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,213 @@
...
+
+            try {
+                throw new StructuredActivityFailure(
+                        FAILURE_DETAILS_JSON_PRINTER.print(failureDetails.toProto()));
+            } catch (InvalidProtocolBufferException serializationException) {
+                LOGGER.log(Level.WARNING,
+                        "Failed to serialize structured failure details; rethrowing the original exception.",
+                        serializationException);
+                throw e;
+            }
```
bachuv, line 89: The fallback only catches `InvalidProtocolBufferException`, but `toProto()` can also throw unchecked exceptions for provider-supplied values, e.g. a null property key produces `NullPointerException`. That replaces the original activity failure instead of preserving it. Could we catch conversion failures around `toProto()` as well, log them, and rethrow e?

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
```diff
@@ -322,10 +331,17 @@ private static Value convertToProtoValue(@Nullable Object obj) {
...
+            for (int index = 0; index < length; index++) {
+                listBuilder.addValues(convertToProtoValue(Array.get(obj, index)));
+            }
+            return Value.newBuilder().setListValue(listBuilder).build();
         } else if (obj instanceof Map) {
             Struct.Builder structBuilder = Struct.newBuilder();
-            for (Map.Entry<String, Object> entry : ((Map<String, Object>) obj).entrySet()) {
-                structBuilder.putFields(entry.getKey(), convertToProtoValue(entry.getValue()));
+            for (Map.Entry<?, ?> entry : ((Map<?, ?>) obj).entrySet()) {
+                structBuilder.putFields((String) entry.getKey(), convertToProtoValue(entry.getValue()));
```
halspang: Why change the `String` to a `?` if we immediately cast to a `String`?

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
nytian: This is used by ActivityMiddleware.java so we have to make it pubic

client/src/main/java/com/microsoft/durabletask/FailureDetails.java
nytian: I just want to remove the unchecked warnings above and thus made this change. You are right its a little bit confusing.. Update it to be more clear.

azurefunctions/src/main/java/com/microsoft/durabletask/azurefunctions/internal/middleware/ActivityMiddleware.java
```diff
@@ -0,0 +1,214 @@
...
+            String serializedFailureDetails;
+            try {
+                serializedFailureDetails = FAILURE_DETAILS_JSON_PRINTER.print(failureDetails.toProto());
+            } catch (InvalidProtocolBufferException | RuntimeException conversionException) {
+                LOGGER.log(Level.WARNING,
+                        "Failed to convert or serialize structured failure details; rethrowing the original exception.",
+                        conversionException);
+                throw e;
+            }
+            throw new StructuredActivityFailure(serializedFailureDetails);
```
bachuv, line 90: Does this JSON survive the Java worker’s exception serialization? The worker’s root-cause formatting appears to prefix the message with the exception type, producing something like  StructuredActivityFailure: {"errorType":...} . If the Durable extension requires the payload to begin with  { , structured parsing will fail and the custom properties/inner failure will be lost. Could we validate this through the actual worker-to-host path and either adjust the wire format or make the host parser handle the prefix? The current unit tests only inspect  getMessage() , so it doesn't look like they cover this boundary.

nytian: Yes The Java worker uses ExceptionUtils.getRootCauseMessage(), so the host receives StructuredActivityFailure: {…}, not raw JSON. The current host parser would reject that prefix. The corresponding change at extension [here ](https://github.com/Azure/azure-functions-durable-extension/pull/3465/changes) will update how extension parse the prefix. And then the e2e test will be passed. I verified that at my end

bachuv (approved):

PR #296, opened 2026-08-07
Add entity tracing support

bachuv:
This pull request introduces enhancements to entity operation trace context propagation and adds comprehensive OpenTelemetry tracing support for Durable Task entities. The main changes ensure that when an entity operation includes a trace context, it is correctly propagated to any signals or orchestrations the entity produces. Additionally, the tracing helper now provides a full set of utilities for emitting and testing entity-related spans, aligning with the .NET SDK schema.

**Entity Operation Trace Context Propagation:**
- Ensured that the `TraceContext` from an incoming entity operation is captured and propagated to outgoing signals and orchestrations, allowing for proper trace linkage in distributed tracing scenarios. (`TaskEntityExecutor.java`)
- Added tests to verify that trace context is correctly propagated (or omitted when not present) for signals and orchestrations initiated by entities. (`TaskEntityExecutorTest.java`)

**OpenTelemetry Tracing Support for Entities:**
- Added entity-specific span constants and helper methods to `TracingHelper`, following the .NET SDK schema. This includes methods for creating span names, starting/ending processing spans, and emitting client/producer spans for entity operations, signals, and orchestration starts.

### Pull request checklist


client/src/main/java/com/microsoft/durabletask/TracingHelper.java
```diff
@@ -476,4 +486,185 @@ static void emitEventSpan(
...
+    }
+
+    /**
+     * Starts a processing span for an entity operation: {@link SpanKind#SERVER} for a call,
+     * {@link SpanKind#CONSUMER} for a signal. Returns {@code null} when the parent context is absent
+     * or invalid, matching the .NET guard that avoids attaching entity spans to an unrelated ambient
+     * trace. The caller makes the span current and later calls {@link #endEntityProcessingSpan}.
+     */
+    @Nullable
+    static Span startEntityProcessingSpan(
```
YunchuWang, line 508: **Must address — wire these helpers into a production path or re-scope the PR.** Repository-wide, the new entity span helpers are called only by `TracingHelperTest`. `DurableTaskGrpcWorker` still invokes `TaskEntityExecutor.execute(...)` directly for both entity request formats, and the client/orchestration/entity-action paths never call the new processing, client, signal-producer, or start-orchestration helpers. As written, registering OpenTelemetry therefore emits no entity spans.

Please either (1) wire processing spans around entity execution, emit client/producer spans at the call/signal/start boundaries, propagate each newly created span's context, and add a production-path trace test; or (2) if the sidecar owns this lifecycle, remove/re-scope these unused Java helpers and document that ownership.

client/src/main/java/com/microsoft/durabletask/TaskEntityExecutor.java
```diff
@@ -121,6 +121,10 @@ EntityBatchResult execute(@Nonnull EntityBatchRequest request) {
             logger.log(Level.FINE, "Executing operation '{0}' (requestId={1}) on entity '{2}'.",
                     new Object[]{operationName, requestId, instanceId});
 
+            // Parent context for any signals/orchestrations this operation produces, so the host can link them.
+            context.setCurrentOperationTraceContext(
+                    opRequest.hasTraceContext() ? opRequest.getTraceContext() : null);
```
YunchuWang: **Must address or explicitly scope — the native orchestration-to-entity path cannot currently provide this context.** In the `ENTITYREQUESTV2` path, `TaskOrchestrationExecutor` creates `EntityOperationCalledEvent` / `EntityOperationSignaledEvent`, but neither protobuf message has a trace-context field. `DurableTaskGrpcWorker` can consequently copy only request ID, operation, and input into `OperationRequest`, making `hasTraceContext()` false for this path. The flattened request also loses whether the operation was a call or signal, so a processing span cannot select `SERVER` versus `CONSUMER` correctly.

Please define which component owns entity span creation and carry the required context/invocation kind through the protocol (with backend coordination), or explicitly limit this PR to propagation for backend-supplied V1 requests. An end-to-end test should cover both an orchestration call and signal.

client/src/main/java/com/microsoft/durabletask/TaskEntityExecutor.java
```diff
@@ -261,6 +271,10 @@ public void signalEntity(
                         .build());
             }
 
+            if (this.currentOperationTraceContext != null) {
```
YunchuWang, line 299: **Should address — also populate `requestTime` here and for `StartNewOrchestrationAction`.** Both protobuf actions already have a `requestTime` field, and the .NET worker shim records `DateTimeOffset.UtcNow` in these exact two paths. The new producer-span helpers accept a start time, so leaving this absent means later wiring either loses the action-creation timestamp (making queue/dispatch latency inaccurate) or risks converting an unset protobuf timestamp to the Unix epoch.

Please set `setRequestTime(toTimestamp(Instant.now()))` on both builders and assert it in the corresponding executor tests.

client/src/main/java/com/microsoft/durabletask/TracingHelper.java
```diff
@@ -476,4 +486,185 @@ static void emitEventSpan(
...
+        if (startTime != null) {
+            spanBuilder.setStartTimestamp(startTime);
+        }
+        Span span = spanBuilder.startSpan();
+        setSpanId(span, syntheticSpanId);
+        if (errorDescription != null) {
+            span.setStatus(StatusCode.ERROR, errorDescription);
+        }
+        if (endTime != null) {
+            span.end(toEpochNanos(endTime), java.util.concurrent.TimeUnit.NANOSECONDS);
```
YunchuWang: **Optional simplification:** OpenTelemetry's `Span` API in the version used here already provides `end(Instant)`. This can be `span.end(endTime)`, allowing the private `toEpochNanos()` helper to be removed. The current conversion is not a practical correctness bug; this only avoids duplicating library functionality.

YunchuWang (commented):
Second-pass review focused on end-to-end entity tracing behavior and parity with the .NET implementation. I found two must-address completeness gaps, one timestamp issue that should be addressed, and one optional simplification. The context propagation logic itself otherwise looks safe and consistent with the .NET worker shim.

client/src/main/java/com/microsoft/durabletask/TaskEntityExecutor.java
YunchuWang: Rechecked at current head `27f6ea2a`: this remains unresolved. The V2 conversion still builds both `OperationRequest` variants without a trace context, and `TaskEntityExecutor` now hardcodes `signal=false` when starting the processing span. Therefore the default native path still cannot link the incoming trace and can never produce a `CONSUMER` processing span for signals. The new test also explicitly expects zero `CLIENT` spans as "deferred pending protocol support". Please either coordinate the protobuf/backend fields needed for context + invocation kind, or explicitly re-scope and track this limitation before merging.

client/src/main/java/com/microsoft/durabletask/TaskEntityExecutor.java
```diff
@@ -261,6 +296,26 @@ public void signalEntity(
...
+                        ? options.getScheduledTime().toString() : null;
+                Span producerSpan = TracingHelper.startEntitySignalProducerSpan(
+                        targetEntityId.getName(),
+                        operationName,
+                        targetEntityId.toString(),
+                        this.entityId.toString(),
+                        this.currentOperationTraceContext,
+                        null,
+                        signalScheduledTime);
+                if (producerSpan != null) {
```
YunchuWang, line 314: **Must address — do not export spans for actions that are later rolled back.** This PRODUCER span is ended immediately inside `signalEntity()` (and similarly in `startNewOrchestration()`), before the surrounding entity operation commits. If the entity signals and then throws, `context.rollback()` correctly removes the pending action, but this already-finished span remains exported and appears to show work that was never sent.

Please defer emission until the action is known to be committed, and add a regression test where an entity queues an action and then throws: the result should contain zero actions **and** export zero corresponding PRODUCER spans. Keep the existing success assertion to prove committed actions still emit exactly once.

client/src/main/java/com/microsoft/durabletask/TaskEntityExecutor.java
```diff
@@ -127,6 +132,23 @@ EntityBatchResult execute(@Nonnull EntityBatchRequest request) {
 
             Instant startTime = Instant.now();
 
+            // Entity processing span parented on the incoming operation trace context;
+            // suppressed when the host emits its own (emitTraceSpans is false).
+            Span processingSpan = this.emitTraceSpans
```
YunchuWang, line 137: **Should address — make the processing span current while user entity code runs.** The span is created and its context is copied manually to Durable actions, but `entity.run(operation)` is never executed under `processingSpan.makeCurrent()`. As a result, spans created by user code or automatic HTTP/database instrumentation inside the entity will not be children of this entity processing span and will appear disconnected.

Please mirror the existing activity execution pattern: open a `Scope` from the processing span around `entity.run(...)`, close it on every success/failure path, and then end the span. Add a test that creates a normal nested span inside an entity operation and verifies its parent is the processing span.

client/src/main/java/com/microsoft/durabletask/TaskEntityExecutor.java
```diff
@@ -261,6 +296,26 @@ public void signalEntity(
                         .build());
             }
 
+            if (this.currentOperationTraceContext != null) {
```
YunchuWang, line 299: **Must address — propagate the PRODUCER span context, not its parent processing context.** The action is stamped with `currentOperationTraceContext` before the PRODUCER span is created, and the new PRODUCER span's context is discarded. A downstream CONSUMER/orchestration span will therefore be parented directly to the processing span, making it a sibling of the PRODUCER instead of its child. The .NET dispatcher starts the producer activity and puts that activity's context on the outgoing message.

After creating the committed PRODUCER span, set the action's `parentTraceContext` from that span's context (for both signal and start-orchestration actions). Add an assertion that the action's propagated span ID equals the exported PRODUCER span ID, not the enclosing SERVER span ID.

client/src/test/java/com/microsoft/durabletask/TaskOrchestrationEntityTracingTest.java
```diff
@@ -0,0 +1,188 @@
...
+        executor.execute(
+                Collections.emptyList(),
+                Arrays.asList(orchestratorStarted(), executionStarted(orchestratorName)),
+                orchestrationContext());
+
+        List<SpanData> spans = spanExporter.getFinishedSpanItems();
+        long producers = spans.stream().filter(s -> s.getKind() == SpanKind.PRODUCER).count();
+        long clients = spans.stream().filter(s -> s.getKind() == SpanKind.CLIENT).count();
+        assertEquals(1L, producers, "signalEntity should emit exactly one PRODUCER span");
+        assertEquals(0L, clients, "callEntity CLIENT span is deferred pending protocol support");
```
YunchuWang, line 186: **Must address or explicitly re-scope — this test currently codifies incomplete call tracing.** `emitEntityCallClientSpan()` still has no production caller, so `callEntity()` never emits the promised CLIENT span. The CLIENT span should cover request-to-response (or timeout/cancellation), use the synthetic span ID that parents the SERVER processing span, and be suppressed during replay.

Please implement and assert the CLIENT span lifecycle rather than permanently expecting zero here. If protocol/backend work makes that impossible in this PR, remove/re-scope the unused helper and state the limitation prominently in the PR description with a tracked follow-up.

YunchuWang (commented):
Follow-up review of the remediation commits. The `requestTime` and `Span.end(Instant)` comments are correctly addressed, and processing/producer wiring has been added. The native V2 protocol limitation remains open (updated in the existing thread). I found three correctness gaps in the new wiring plus the still-deferred call/external-client coverage below. This is a comment-only review; it does not change the PR review state.

Also, `GrpcDurableEntityClient.signalEntity()` remains unchanged and still captures only the ambient context without emitting the new PRODUCER span, so external-client signals are not covered by the advertised entity tracing support. Please wire that path or explicitly document/re-scope it as deferred.

YunchuWang:
Follow-up at current head `27f6ea2a`: there have been no code changes since the Aug 18 review, so the latest findings are still reproducible and should remain open.

Correctly addressed and safe to keep resolved:
- `requestTime` is populated and tested.
- `Span.end(Instant)` replaced the manual epoch-nanos conversion.

Still unresolved:
- Native `ENTITYREQUESTV2` cannot carry trace context or call-vs-signal kind, so signal processing cannot become `CONSUMER` and the default path cannot preserve the intended trace hierarchy: https://github.com/microsoft/durabletask-java/pull/296#discussion_r3806587419
- Rolled-back actions still export phantom PRODUCER spans: https://github.com/microsoft/durabletask-java/pull/296#discussion_r3806591271
- The processing span is not made current around user entity code: https://github.com/microsoft/durabletask-java/pull/296#discussion_r3806591275
- Outgoing actions propagate the processing context rather than the PRODUCER span context: https://github.com/microsoft/durabletask-java/pull/296#discussion_r3806591276
- Entity calls still intentionally emit zero CLIENT spans, and the external `GrpcDurableEntityClient.signalEntity()` path is not wired to the new producer helper: https://github.com/microsoft/durabletask-java/pull/296#discussion_r3806591281

The current CI is green, but it does not exercise these failure and end-to-end hierarchy cases. I would keep the PR in changes-requested territory until these are fixed or the PR is explicitly re-scoped with tracked follow-ups.

bachuv:
Thanks @YunchuWang, here are the updates from my changes:

- Removed `TracingHelper.endEntityProcessingSpan(processingSpan, null)` because the processing span is owned by the DTFx dispatcher.
- For trace-bearing V1 requests, DTFx creates the correct `SERVER` or `CONSUMER` span and passes its context through `OperationRequest.traceContext`.
- Java now makes that context current while running entity code, allowing nested user and PRODUCER spans to inherit it.
- DTFx ends the original processing span after receiving the operation result.
- Creating and ending another Java span produced a duplicate hierarchy and incorrectly classified every operation as a call (`SERVER`).
- Native V2 processing spans remain deferred because the protocol does not carry trace context or call-versus-signal kind.

YunchuWang (approved):
Reviewed at ad9da817. The code-level findings are addressed: entity execution makes the incoming dispatcher-owned context current, producer spans are emitted only for committed actions, committed actions propagate the producer context, external-client signals are covered, and the prior timestamp/helper issues are fixed. Native V2 and callEntity tracing remain an explicit protocol/backend limitation and can be completed in coordinated follow-up work.

PR #304, opened 2026-09-22
Fix large-payload recovery and export-history parity with .NET

bachuv:

Follow-up to #280 and #293. Addresses three behavioral differences from .NET that existed when the Java large-payload and export-history features were merged.

### Pull request checklist

### Additional information

**Changes**

* Recreate deleted payload containers and retry new uploads once. Generation-aware cache invalidation prevents stale concurrent failures from invalidating a container another upload has already recreated.
* Preserve sub-millisecond timestamps in orchestration history and client metadata so exported timestamps and blob filenames do not diverge from .NET due to truncation. Existing orchestration replay timestamp behavior remains unchanged.
* Export entity operations and locks using .NET-compatible `EventSent`/`EventRaised` messages, including parent context, scheduled operations, nested failure details, and JSON escaping. The public history API continues returning typed Java entity events.

**Validation**

* Added regression coverage for container recovery, concurrent failures, timestamp precision, and export formatting.
* Added test-only golden fixtures generated using .NET history conversion and DurableTask.Core 3.9.0.
* All 185 scoped tests passed: 43 client, 79 large-payload, and 63 export-history tests.

**Compatibility**

The preview entity-export format changes from Java-native entity objects to the .NET-compatible representation. Re-exporting instances previously named using millisecond-truncated timestamps may create different blob names; existing blobs are not renamed or removed.


exporthistory/src/main/java/com/microsoft/durabletask/exporthistory/HistoryEventSerializer.java
```diff
@@ -382,22 +398,149 @@ private static String formatInstant(Instant t) {
...
+        if (failure == null) {
+            return null;
+        }
+        Map<String, Object> result = new LinkedHashMap<>();
+        result.put("ErrorType", failure.getErrorType());
+        result.put("ErrorMessage", failure.getErrorMessage());
+        result.put("StackTrace", failure.getStackTrace());
+        result.put("InnerFailure", entityFailureMap(failure.getInnerFailure()));
+        result.put("IsNonRetriable", failure.isNonRetriable());
+        result.put("Properties", failure.getProperties() == null ? Collections.emptyMap() : failure.getProperties());
```
YunchuWang: **[P2] Decode date-valued failure properties in the export projection**

Copying `failureDetails.getProperties()` directly preserves the `dt:`/`dto:` transport prefixes that .NET uses for date-valued properties. Java's protobuf conversion leaves those strings unchanged, whereas the reference [.NET conversion recursively decodes them](https://github.com/microsoft/durabletask-dotnet/blob/cfe26800bbb693d3d4ddaa74ca7a5c3b457a9dc1/src/Shared/Grpc/ProtoUtils.cs#L1057-L1095) before Newtonsoft serializes the entity response. For example, a streamed `EntityOperationFailed` property containing `dt:2026-06-30T12:00:00.0000000Z` exports with that prefix here, but as `2026-06-30T12:00:00Z` in .NET. Consumers parsing the reference ISO date value therefore receive a different, non-date string.

Please add an export-local recursive property projection matching the reference date decoding/formatting, preserving ordinary and invalid tagged strings, and extend the streamed-history parity cases to include `dt:`, `dto:`, nested maps/lists, and inner failures. This can fix the new export representation without changing the public Java `FailureDetails` contract.

YunchuWang (changes requested):
Requesting changes for one remaining value-level mismatch in the new .NET-compatible entity-failure export. Reviewed commit `5ec000e88240e91f62d19e3769662a800baf4974`.

